PR #22, opened 2017-10-17
Add .pre-commit-hooks.yaml so this can be used with pre-commit

brainsik:
Added necessary file for using this with [pre-commit](http://pre-commit.com/).


DavidAnson:
Obsoleted by https://github.com/igorshubovych/markdownlint-cli/commit/ac321575cbeacc4e4931bfb2fd281215afb0d872 which does the same thing.

Thanks!

PR #29, opened 2017-11-10
fix #23 - use console error; process.exit alone; test

### LINKED ISSUE #23: End reports with an endline
image

The report should have one or two endlines after it so the next `npm ERR!` is on its own line.

davidosomething:
- Use console.error (still outputs to stderr)
- Simplify process.exit call
- Add "invalid" script for testing; test with regex newline where an npm run error would output on the same line

### DIFF
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -61,21 +61,28 @@ function lint(lintFiles, config) {
 
 function printResult(lintResult, hasErrors) {
   if (hasErrors) {
-    process.stderr.write(lintResult.toString(), process.exit.bind(process, 1));
-  } else {
-    var result = lintResult.toString();
-    if (result) {
-      console.log(result);
-    }
+    console.error(lintResult.toString());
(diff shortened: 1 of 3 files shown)


davidosomething:
Hey @igorshubovych  can you npm publish after this one?

DavidAnson:
The change to exit behavior *looks* right, but is documented as being wrong:
https://nodejs.org/dist/latest-v8.x/docs/api/process.html#process_process_exit_code

Once fixed, I could accept the PR and publish an update over the weekend.

davidosomething:
good to know, i will switch to process.exitCode and natural termination

DavidAnson:
Great, thank you! When you’re making that change, could you please add a comment there referencing the documentation so that future generations know why it’s been done this way? I only barely remembered the reason this time and might forget next time. :)

davidosomething:
ready!

DavidAnson:
Beautiful.

PR #30, opened 2017-12-21
fixes globbing issues; add support for *.markdown in dirs; closes #20

### LINKED ISSUE #20: Support cross-platform glob
It would be nice to support cross-platform glob like [prettier](https://github.com/prettier/prettier#cli).

Currently markdownlint-cli supports directory.
But if directory has node_modules directory, markdownlint-cli also search in it.
To avoid that, I want to use glob like this:

```
# ./ has node_modules directory and *.md file exists in ./ and ./docs
markdownlint './*.md' './docs/**/*.md'
```

boneskull:
- globbing should work in cross-platform manner in `package.json`; e.g.
  `markdownlint "docs/**.md"`
- add tests + fixtures


DavidAnson:
This looks reasonable, thanks! I won’t be able to merge for a week or two, but will do so when I am able.

In the meantime, a couple of questions:

+ Why remove a test case? Doesn’t it still run/pass? If not, wouldn’t that be a problem for users?
+ Could you please add a failure file to the glob testing to be sure it is reported correctly?
+ What do you think about putting the new return statement inside the catch block?

Thanks again!

boneskull:
> Why remove a test case? Doesn’t it still run/pass? If not, wouldn’t that be a problem for users?

whoops :smile:

> Could you please add a failure file to the glob testing to be sure it is reported correctly?

sure

> What do you think about putting the new return statement inside the catch block?

I can rearrange these, yes

boneskull:
I've made the requested changes

boneskull:
I added some more fixtures to convince myself it's working.  IMO the assertions are flimsy (because there are other reasons than linting failures that they could occur), but unless this tool can provide machine-readable output, it's about the best we can do

DavidAnson:
Looks great, thanks! Agree about the assertions, that would be a good thing to improve, but unrelated to this change. A low-cost way to be a bit better might be to add some  string matching to be sure the expected file names are present in the output.

boneskull:
Just wanted to ping since it's been a week.  :wink:  

This impacts mochajs/mocha#3173

DavidAnson:
Sorry for the delay - it is holiday season in my part of the world. I should have time for this soon!

DavidAnson:
BTW: Can you explain the comment there “which is dumb, but we're stuck with it for now”? Is that anything to do with this project, or something else? Could it be fixed here?

DavidAnson:
I published a new version with your change just now. Please let me know if that doesn’t do what you need!

boneskull:
@DavidAnson Thanks!!  I'll use this and delete my fork

boneskull:
> BTW: Can you explain the comment there “which is dumb, but we're stuck with it for now”? Is that anything to do with this project, or something else? Could it be fixed here?

No, this is in reference to Mocha's use of a `Makefile`

Issue #36, opened 2018-05-10
WARN notice [SECURITY] deep-extend has 1 low vulnerability


dex4er:
markdownlint-cli uses deep-extend@0.4.2 which is vulnerable and npm reports it:

```
npm WARN notice [SECURITY] deep-extend has 1 low vulnerability. Go here for more details: https://nodesecurity.io/advisories?search=deep-extend&version=0.4.2 - Run `npm i npm@latest -g` to upgrade your npm version, and then `npm audit` to get more info.
```

See https://nodesecurity.io/advisories?search=deep-extend&version=0.4.2

PR #37, opened 2018-05-11
Update deep-extend and rc

mroderick:
In order to avoid security warning https://nodesecurity.io/advisories/612

You can see the security warnings by running `npm audit` (with `npm@6`)


DavidAnson:
Thanks!

Could you please update the commit to say it fixes #36?

Also, the warning you link to does not reference `rc`, is that also affected?

And did you review the list of breaking changes for `rc`? (I already did for `deep-extend`.)

PR #40, opened 2018-06-14
Add support of custom rules

VeselovAlex:
Added a way to pass custom rule scripts to linter.
* Use `-r` or `--rules` arg with path to js file following

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,13 +19,13 @@ $ markdownlint --help
 
   Options:
 
-    -h, --help                          output usage information
-    -V, --version                       output the version number
-    -s, --stdin                         read from STDIN (no files)
-    -o, --output [outputFile]           write issues to file (no console)
-    -c, --config [configFile]           configuration file (JSON or YAML)
-    -r, --rules  [file|directory|glob]  custom rule files
-    -i, --ignore [file|directory|glob]  files to ignore/exclude
+    -h, --help                                  output usage information
+    -V, --version                               output the version number
+    -s, --stdin                                 read from STDIN (no files)
+    -o, --output [outputFile]                   write issues to file (no console)
+    -c, --config [configFile]                   configuration file (JSON or YAML)
+    -i, --ignore [file|directory|glob]          files to ignore/exclude
+    -r, --rules  [file|directory|glob|package]  custom rule files
 ```
 
 ## Configuration
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ $ markdownlint --help
     -s, --stdin                         read from STDIN (no files)
     -o, --output [outputFile]           write issues to file (no console)
     -c, --config [configFile]           configuration file (JSON or YAML)
-    -r, --rules [file]                  custom rule files
+    -r, --rules  [file|directory|glob]  custom rule files
     -i, --ignore [file|directory|glob]  files to ignore/exclude
 ```
 
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ $ markdownlint --help
     -s, --stdin                         read from STDIN (no files)
     -o, --output [outputFile]           write issues to file (no console)
     -c, --config [configFile]           configuration file (JSON or YAML)
+    -r, --rules [file]                  custom rule files
     -i, --ignore [file|directory|glob]  files to ignore/exclude
 ```
 
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -41,14 +41,22 @@ function readConfiguration(args) {
   return config;
 }
 
-function prepareFileList(files) {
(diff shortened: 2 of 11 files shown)


DavidAnson:
Wow, thank you!! Looking quickly on my phone, your changes make sense. But I will need to review more carefully this evening or weekend. One thing that would be nice to add before then are test cases. (And I’ll ask you to squash the changes into one.)

README.md
```diff
@@ -24,6 +24,7 @@ $ markdownlint --help
     -s, --stdin                         read from STDIN (no files)
     -o, --output [outputFile]           write issues to file (no console)
     -c, --config [configFile]           configuration file (JSON or YAML)
+    -r, --rules  [file|directory|glob]  custom rule files
```
DavidAnson: Or packages? I think what you did for files will also support packages, so let’s say that here.

markdownlint.js
```diff
@@ -41,14 +41,22 @@ function readConfiguration(args) {
   return config;
 }
 
-function prepareFileList(files) {
+function prepareFileList(files, fileExtensions) {
   var globOptions = {
     nodir: true
   };
+  var extensionGlobPart = '*';
+  if (fileExtensions && fileExtensions.length > 0) {
```
DavidAnson: This is an internal function, so is the parameter validation necessary?

markdownlint.js
```diff
@@ -41,14 +41,22 @@ function readConfiguration(args) {
...
+    if (fileExtensions.length === 1) {
+      extensionGlobPart += '.' + fileExtensions[0];
+    } else {
+      extensionGlobPart += '.{' + fileExtensions.join(',') + '}';
```
DavidAnson: Will this syntax work for just one item? (I’d expect it to, but it needs testing.) If it does, the if/else can be removed.

markdownlint.js
```diff
@@ -117,12 +125,23 @@ program
   .option('-s, --stdin', 'read from STDIN (no files)')
   .option('-o, --output [outputFile]', 'write issues to file (no console)')
   .option('-c, --config [configFile]', 'configuration file (JSON or YAML)')
+  .option('-r, --rules  [file|directory|glob]', 'custom rule files', concatArray, [])
```
DavidAnson: I like how you lined up the brackets here.

markdownlint.js
```diff
@@ -117,12 +125,23 @@ program
...
 
 program.parse(process.argv);
 
-var files = prepareFileList(program.args);
-var ignores = prepareFileList(program.ignore);
+function loadCustomRuleFromFile(filepath) {
+  try {
+    return require(filepath.absolute);
+  } catch (err) {
+    console.error('Cannot load custom rule form ' + filepath.original + ': ' + err.message);
```
DavidAnson: form -> from

markdownlint.js
```diff
@@ -117,12 +125,23 @@ program
...
+    process.exitCode = 3;
```
DavidAnson: I think we need to return or be sure not to run the code below some other way?

markdownlint.js
```diff
@@ -117,12 +125,23 @@ program
...
+  }
+}
+
+var files = prepareFileList(program.args, ['md', 'markdown']);
+var ignores = prepareFileList(program.ignore, ['md', 'markdown']);
+var customRules = prepareFileList(program.rules, ['js']).map(loadCustomRuleFromFile);
```
DavidAnson: I like how you did this.

README.md
DavidAnson: Wait, maybe that is wrong. The new code searches for all .js files, then require-s them all separately. So packages are not supported. What if it was used like this instead: —rules file.js dir/package

DavidAnson: Basically, just require each of the things that was passed to —rules.

VeselovAlex:
`package.json` modifications:
* Test file directory declaration moved from scripts to `ava` parameters, ignoring `custom-rules` directory
* Add `test-rule-package` from `custom-rules` for testing

DavidAnson:
Looks awesome! I’ll review in more detail in a day or two. Thanks so much!

markdownlint.js
```diff
@@ -117,12 +124,37 @@ program
...
+function loadCustomRuleFromFile(filepath) {
+  try {
+    return require(filepath.absolute);
+  } catch (err) {
+    console.error('Cannot load custom rule from ' + filepath.original + ': ' + err.message);
+    process.exit(3);
+  }
+}
+
+function tryResolvePath(filepath) {
```
DavidAnson: Nice, I like this approach!

DavidAnson:
Thank you again! I will publish a new version in a week or so.

Issue #42, opened 2018-07-15
Allow including a custom rule via package name


JoshuaKGoldberg:
This works right now:

```cmd
markdownlint --config .markdownlint.json --rules ./node_modules/sentences-per-line/index.js README.md
```

...but this does not:

```cmd
markdownlint --config .markdownlint.json --rules sentences-per-line README.md
```

Relevant function: `prepareFileList`

DavidAnson:
I think this should work if the custom rule’s `package.json` is correct. Here is the test case for it: https://github.com/igorshubovych/markdownlint-cli/blob/master/test/test.js#L381

Can you share the `package.json` for your scenario?

JoshuaKGoldberg:
Sure: https://github.com/JoshuaKGoldberg/sentences-per-line/blob/6a99e70507129baa80bbca4f2b754ee972b18789/package.json

DavidAnson:
Thanks, that looks okay from my phone. Will need to investigate.

DavidAnson:
Fixed by @VeselovAlex

PR #43, opened 2018-07-23
Fix bad module resolution (#42)

VeselovAlex:
Replace `require.resolve('packagename')`, which looks for packages relative to markdownlint-cli root, with `require.resolve('packagename', {paths})`, where paths are calculated relative to working directory


VeselovAlex:
Why does is fail on AppVeyor? Looks like it does not copy package from `test\custom-rules\relative-to-cwd\node_modules`

DavidAnson:
AppVeyor uses Windows, if that is not what you use, maybe there is a path issue there?

Regarding the _ APIs, aren’t those undocumented and unsupported? That worries me, as does the Node 8 requirement. I would love if we could figure out a way to get the right behavior on Node 6 with public APIs.

VeselovAlex:
Node 8 is not required. `require.resolve.paths` and `require.resolve(module, paths)` are added to public API in Node 8, so we can use them to search packages relative to working directory. For Node 6 we can use `Module._resolveFilename(filepath, {paths: paths})` as a fallback.

To replace `Module._nodeModulePaths(process.cwd())` we can create custom function to build node resolution path list, but should we? This method is available even in 0.10.25 and was not modified for at least 2 years, so it seems to be stable enough

VeselovAlex:
Tests fixed, seems to an issue with `execa` on Windows

DavidAnson:
Yay!!

(Did you mean to leave that console.log in the code?)

DavidAnson:
I should be able to accept this PR tonight. I’ll also update the markdownlint dependency and investigate the other open issue. Once done, I’ll publish a new version. Thanks for the help!

PR #47, opened 2018-12-26
Added Dockerfile for the markdownlint-cli

andmos:
Hi 👋 
Found myself in the need for a `Dockerfile` for the project, as I use it as a part of a containerized CI pipeline. Thought I'd share the file with the project if desired 😃

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+.node_repl_history
+.vscode/
+.idea/
+test/
+.editorconfig
+.travis.yml
+appveyor.yml
+node_modules\
+dist\
+node_modules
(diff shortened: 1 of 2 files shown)


DavidAnson:
Thank you! I don’t use Docker, so if someone else who does is able to review these changes, that would be great. My one comment would be that there are a lot of lines in the “ignore” file that aren’t relevant. My preference would be to remove them, but if this is a standard template, that seems reasonable.

hansehe:
Hi, I know Docker! The dockerfile looks fine and works, and the .dockerignore file was updated recently to not have too many irrelevant ignore lines, so I would suggest approving this PR.

DavidAnson:
Thank you both!

PR #48, opened 2019-01-10
Configuration is stored in JSON, YAML or INI files

hugovk:
Two other places in the README say "JSON or YAML", so should this also say that instead of "JSON or INI"?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ $ markdownlint --help
 
 `markdownlint-cli` reuses [the rules][rules] from `markdownlint` package.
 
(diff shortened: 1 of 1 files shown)


DavidAnson:
This project uses the `RC` package, so I believe it is correct to include INI there: https://www.npmjs.com/package/rc

But I think you are right that it would be good to include YAML as well.

The other two references you mention relate the file that is passed to the `—config` switch and in that case both references are already correct as they are now.

Could you please update this PR?

hugovk:
Updated!

PR #64, opened 2019-10-01
Improve format of error message

programmer04:
Example error message in the previous format:

```
docs/glossary.mdx: 16: MD004/ul-style Unordered list style [Expected: dash; Actual: asterisk]
```
When user clicks on file path from above message in an integrated terminal for example in VS Code was moved to file, but not to a specific line with an error.

After my update message looks like below:

```
docs/glossary.mdx:16 MD004/ul-style Unordered list style [Expected: dash; Actual: asterisk]
```

now user after clicking on file path is moved directly to file and line where an error occurred.

I think this is more comfortable in usage. What do you think?

### DIFF
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -127,7 +127,7 @@ function printResult(lintResult) {
         a.names.localeCompare(b.names) || a.description.localeCompare(b.description);
     });
     lintResultString = results.map(function (result) {
-      return result.file + ': ' + result.lineNumber + ': ' + result.names + ' ' + result.description;
+      return result.file + ':' + result.lineNumber + ' ' + result.names + ' ' + result.description;
     }).join('\n');
     // Note: process.exit(1) will end abruptly, interrupting asynchronous IO
     // streams (e.g., when the output is being piped). Just set the exit code
--- a/test/test.js
+++ b/test/test.js
@@ -5,7 +5,7 @@ import path from 'path';
 import test from 'ava';
 import execa from 'execa';
 
-const errorPattern = /(\.md|\.markdown|stdin): \d+: MD\d{3}/gm;
+const errorPattern = /(\.md|\.markdown|stdin):\d+ MD\d{3}/gm;
 
 process.chdir('./test');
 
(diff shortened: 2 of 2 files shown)


DavidAnson:
Makes sense! I’ll be releasing new bits soon and will review your change in more detail then. Thank you!

In the meanwhile, could you please confirm this is a generally relevant pattern and that we are not just improving the experience for VS Code users with this change?

programmer04:
Thanks for quick reply. It's a common pattern, Visual Studios support it. IDEs from JetBrains (IntelliJ, PyCharm, WebStorm etc.) are working on supporting navigating by clicking in terminal (see [feature request](https://youtrack.jetbrains.com/issue/IDEA-118566)) but when you copy path in this format `docs/glossary.mdx:16` from terminal and paste into file navigator (`Ctrl`+`Shift`+`N`) it moves to the specified file and line automatically (`docs/glossary.mdx: 16:` doesn't work). These editors have big market share according to [StackOverflow Survey 2019](https://insights.stackoverflow.com/survey/2019#development-environments-and-tools). Moreover many other linters and compilers follow this syntax too (e.g. flake or go compiler).

DavidAnson:
Great, thanks! Do they also support file.txt:12:34 for column 34 of line 12? Many rules already include range info with errors and we could enhance this further. Not in this commit, but maybe one day.

nschonni:
Had trouble finding the real documentation, but the CLI seems to indicate that is the correct pattern https://github.com/microsoft/vscode-docs/blob/a02163963e53c1ef110c9d18261ab2665edec01c/docs/getstarted/tips-and-tricks.md#command-line

programmer04:
This pattern `file.txt:12:34` is also supported by editors. As @nschonni mentioned this format `file:line[:character]` (section `[:character]` is optional) is a quite standard way of doing it. Enhancing support to character will be nice as a future improvement. It is applied only to some rules so it is more challenging to implement. Probably let's better do it in the next PR.

DavidAnson:
Thanks!

PR #84, opened 2020-04-15
Bump markdownlint to current version (0.20.1)

polarweasel:
Is there anything else required to use Markdownlint v0.20.1?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4738,9 +4738,9 @@
       }
     },
     "markdownlint": {
-      "version": "0.19.0",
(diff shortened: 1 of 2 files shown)


DavidAnson:
Looks like you need to update `package-lock.json` to pass tests?

DavidAnson:
Please also bump the version of `markdownlint-rule-helpers`.

polarweasel:
Thanks! Will do.

polarweasel:
@DavidAnson All set, I think... And thanks for the markdownlint tooling! I use it every day at PubNub in VS Code and the command line, and it's about to get integrated into some CI toolchains. :)

DavidAnson:
Thanks!

PR #86, opened 2020-04-29
bumping Dockerfile to node 12

andmos:
The old Node 8 baseimage was a bad fit, bumping to 12 (LTS version of Node)


DavidAnson:
Thanks!

Issue #85, opened 2020-04-15
Add support to javascript config file format (.markdownlintrc.js)


alejandroclaro:
It would be nice to add support to javascript config files like eslint and stylelint support.

I have a use case where we share the configuration files using a NPM package. With eslint and stylelint it is very elegant to be able to import the configuration using the NodeJS package resolution algorithm. Something like this:

.markdownlintrc.js
``` javascript
const config = require('@my-company/configuration/markdownlintrc');

module.exports = config;
```

and the configuration file in the `configuration` package to be something like this:

packages/configuration/markdownlintrc.js
``` javascript
module.exports = {
  "header-increment": true,
  "first-header-h1": true,
  "header-style": {
    "style": "atx"
  }
}
```
instead we have to do this in our packages:

package.json
``` json
"scripts": {
  "lint:md": "markdownlint --ignore ./node_modules/ --config node_modules/@my-company/configuration/markdownlint.json *.md packages/**/*.md"
}
```
but would be nice to be able to do this instead:

package.json
``` json
"scripts": {
  "lint:md": "markdownlint --config .markdownlintrc.js *.md packages/**/*.md"
}
```

PR #88, opened 2020-05-01
Add support for javascript files and package.json configuration

alejandroclaro:
These changes add support for javascript configuration files and package.json embedded configuration. This also makes the configuration file lookup similar to other linters like eslint and stylelint.

BREAKING CHANGE: Replace 'rc' package by 'cosmiconfig'. This changes how configuration files are located: Stops at the first configuration found, instead of finding all that can be found up the directory tree and merge them automatically.

This solves #85

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ markdownlint --help
     -f, --fix                                   fix basic errors (does not work with STDIN)
     -s, --stdin                                 read from STDIN (does not work with files)
     -o, --output [outputFile]                   write issues to file (no console)
-    -c, --config [configFile]                   configuration file (JSON, JSONC, or YAML)
+    -c, --config [configFile]                   configuration file (JS, JSON, JSONC, or YAML)
     -i, --ignore [file|directory|glob]          file(s) to ignore/exclude
     -p, --ignore-path [file]                    path to file with ignore pattern(s)
     -r, --rules  [file|directory|glob|package]  custom rule files
@@ -67,7 +67,7 @@ Because this option makes changes to the input files, it is good to make a backu
 
 `markdownlint-cli` reuses [the rules][rules] from `markdownlint` package.
 
-Configuration is stored in JSON, JSONC, YAML, or INI files in the same [config format][config].
+Configuration is stored in JS, JSON, JSONC, YAML, or INI files in the same [config format][config].
 
 The example of configuration file:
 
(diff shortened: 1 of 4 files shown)


DavidAnson:
Believe it or not, I was going to implement this tonight (or very soon). So I have some established opinions that are different than your approach.

One is that JS config should only be used when explicitly specified - because that involves arbitrary code execution.

The other is to generally avoid breaking changes when not necessary - can you please explain that part of the change?

Thanks!

DavidAnson:
To be clear, `rc` was not my personal choice and I don’t love everything about it - but it’s what we have and people have taken dependencies on it.

DavidAnson:
I’ve spent a little more time looking at the changes you propose and I must say that I am impressed that all the tests continue to pass without modification! The `cosmiconfig` library seems pretty nice.

However, it seems as I look closer that this change removes the ability for one config file to “extends“ another. That was a specific request and is used in practice, so would be another aspect of this being a breaking change that I have concerns about.

There are two other small commits on my list first (one half done), but I hope to have time later today to write and share a private branch to show the change I had in mind.

DavidAnson:
Here's what I had in mind (tests and README to-be-done): https://github.com/igorshubovych/markdownlint-cli/commit/e30b4a93dae2d624c3f7dca02eebd2c5c80259d1

alejandroclaro:
Hi @DavidAnson understood. I did this way based on the stylelint and eslint implementations because I thought that consistency with those linters could be something good.

Stylelint is implemented using cosmiconfig, and eslint has its own implementation, but both behave identically. The preference of the order of the file lookup is also based on these linters, so no surprise is expected if you know them.

I understand that the mechanisms to “extends“ configuration is based on 'rc' package, but didn't know if this was actually something desirable. I like more how stylelint and eslint do it because it is more explicit. However, I completely understand you prefer this mechanism and keep backward compatibility.

Please discard the Pull Request and lets do it the way you propose.

DavidAnson:
OK, cool, thank you for the thoughtful reply! I’ll go with the change I sketched out, but wanted to let you know I like the idea of consistency as you presented and you’ve got me thinking about a new/related project that might use that approach!

DavidAnson:
Closing per discussion - thank you again for the contribution!

alejandroclaro:
Thanks @DavidAnson !

PR #91, opened 2020-05-06
chore: Add Canary in the Gold Mine build

nschonni:
Node uses these to check popular repositories so that breaking changes can be caught early

### DIFF
--- a/.github/workflows/citgm.yml
+++ b/.github/workflows/citgm.yml
@@ -1,0 +1,33 @@
+name: Canary in the Gold Mine build
+
+on: [push, pull_request]
+
+jobs:
+  citgm:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        include:
+        - repo: dotnet/docs
+          command: markdownlint "**/*.md" -i "samples/**/*.md"
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - run: npm ci
+
+    - name: Link Markdownlint globally
+      run: sudo npm link
+
+    - name: Checkout ${{ matrix.repo }}
+      uses: actions/checkout@v2
+      with:
+        repository: ${{ matrix.repo }}
+        path: ${{ matrix.repo }}
+
+    - name: Test ${{ matrix.repo }}
+      run: |
+        cd ${{ matrix.repo }}
+        ${{ matrix.command }}


DavidAnson:
This is a great idea! Do you have any pointers to docs or examples of this? I’d like to review that before accepting. I have a few questions, but they may be answered upon review.

nschonni:
Here is the node project https://github.com/nodejs/citgm
They have a lookup file https://github.com/nodejs/citgm/blob/master/lib/lookup.json and some more advanced config for "flaky" repos

DavidAnson:
One key difference between Node.js and this project: it's expected that rules in this project will sometimes capture new violations in existing code. Therefore it's possible that a correct change to `markdownlint` would break one or more repositories covered by this workflow and cause it to fail. What's the right way to handle that in the relevant commit?

nschonni:
I wouldn't expect these to be blocking for a PR to land, but would give you a view of what might be breaking (correctly or not) in a new release. It would show you in those repos what files are getting new warnings, and you could possibly open Issues there ahead of time if it looks like they are actually legitimate new issues vs. regressions

nschonni:
Not sure what other repos to add. Did a quick search https://github.com/search?o=desc&q=filename%3A.markdownlint.json&s=indexed&type=Code and maybe look at some of the others like https://github.com/gitlabhq/gitlabhq

DavidAnson:
Where I really need this kind of support is the library project itself. Once a regression makes it to the CLI, it’s already out in the wild. I’m partway through doing something similar to this in that project and will use what I learn to inform a review of what you’ve done here. Your suggestion of GitLabHQ is a good one. I’m thinking to pick five repositories based on the feedback that came up recently and use them to help validate the fixes.

nschonni:
If you're looking at doing it the library, here is what I can see:
1. you could either take this approach and write a light wrapper
1. you could clone this project and link the master as the dependency, then run the tests. We do something like that for libsass and sassc as the CLI

DavidAnson:
I wanted something that was easy to run locally and could be updated as part of a commit to exclude files as necessary according to the changes in that commit without blocking on the external repo. Here’s what I came up with, I think it’s your option 1: https://github.com/DavidAnson/markdownlint/compare/testrepos

DavidAnson:
(The new action is correctly failing because one of the repos it tests triggers a regression I haven’t fixed yet.)

nschonni:
Yup, that looks like a good approach 👍 . I might go with a config/lookup JSON file and loop over it vs. the individual tests per repo, but that's just a style preference

DavidAnson:
Having been inspired by this PR and added something similar to the core library project, I’m not sure it’s needed here as well. The docs repository is quite large, so I didn’t use it as part of every CI run - but I hooked up three others of modest size and added provisions to test with the docs repo locally. What do you think?

nschonni:
Probably makes sense to skip the large repositories on every PR, but maybe a scheduled job might make sense for some of the other bigger repos.
Not sure if it still makes sense here, but the one reason I could still see running some in this repo would be to see if there are regressions outside of the core library when mixing in the extras of the CLI/VS Code. EX: ignorefile, nested config files

DavidAnson:
I’d expect things like ignore file and nested config to be sufficiently tested within this repo - or else I will add coverage for gaps. The challenge with rule behavior changes is the overwhelming number of scenarios and surprises hiding in other repos. That’s where I see the most value from outsourcing coverage like this.

nschonni:
@DavidAnson fair enough. I've opened up https://github.com/DavidAnson/markdownlint/pull/290 to run the large repos as a cron job instead

PR #103, opened 2020-06-25
JUnit

### LINKED ISSUE #58: [feature request] junit test reporter
Hi, thank you for this convenient cli!

I'd like to suggest a feature which might be cool especially in CI environment.
If markdownlint-cli supports junit-style(de facto fa) reporter, it'd be so comfortable to check the lint results in CI. 
Other popular test runner and linters (e.g. jest, eslint) does support the feature as well.

How do you feel about this?

FISHMANPET:
Fixes #58 
As requested, a PR for my changes.  I added a little bit since I wrote the comment last night, in 
95cb1ff.  

I added some time and timestamps because Azure Devops was complaining about not having them, and while Azure Devops could handle it I suspect other CIs might not.  Since there are no time or duration values in the `lintResult` object that was already being constructed (and it doesn't look like the base markdownlint library produces any either) I just put in some simple defaults: The timestamp for the test suite will be the time the report is generated, the time for each testcase will just be 1 millisecond (spec is for time in seconds so 0.001 seconds), and the total time for the testcase will be the elapsed time between when the report generation started and when it finishes (which will only be a couple of milliseconds most likely).

I looked a bit at the `junit-report-builder` code to see what all options that could be used, and discovered failures could have a "type" in addition to a "message" so I added the `name` as the type for each failure.

I don't know that there's an official formal JUnit schema, but Azure Devops links to [this one](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd) I'm not very good at reading xml schema documents but it looks like there's some fields that are required according to the schema but that this module doesn't support.  I believe I've used every test suite and test case attribute that exists in the module, and based on its popularity I'm guessing what it produces is acceptable to most if not all CI tools, but if someone would ever report an issue about this generating something that their CI tool can't read, it would probably be an upstream issue for that module to support it before this project can support it.

I appreciated the pre-commit linter (once I realized it was linter errors that were preventing me from committing).  I split up the test suite and test case across multiple lines since they were pretty long, I'm pretty sure there will be other style changes you'd like to see made.

And testing, if you point me in a direction I can try to implement some kind of testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,12 @@ markdownlint --help
     -V, --version                               output the version number
     -f, --fix                                   fix basic errors (does not work with STDIN)
     -s, --stdin                                 read from STDIN (does not work with files)
-    -o, --output [outputFile]                   write issues to file (no console)
-    -c, --config [configFile]                   configuration file (JSON, JSONC, JS, or YAML)
-    -i, --ignore [file|directory|glob]          file(s) to ignore/exclude
-    -p, --ignore-path [file]                    path to file with ignore pattern(s)
-    -r, --rules  [file|directory|glob|package]  custom rule files
+    -o, --output <outputFile>                   write issues to file (no console)
+    -j, --junit <junitFile>                     write issues to file in JUnit format and to the console
+    -c, --config <configFile>                   configuration file (JSON, JSONC, JS, or YAML)
+    -i, --ignore <file|directory|glob>          file(s) to ignore/exclude
+    -p, --ignore-path <file>                    path to file with ignore pattern(s)
+    -r, --rules  <file|directory|glob|package>  custom rule files
 ```
 
 ### Globbing
@@ -62,6 +63,13 @@ When the `--fix` option is specified, `markdownlint-cli` tries to apply all fixe
 Because this option makes changes to the input files, it is good to make a backup first or work with files under source control so any unwanted changes can be undone.
 
 > Because not all rules include fix information when reporting errors, fixes may overlap, and not all errors are fixable, `--fix` will not usually address all errors.
+
+### JUnit report
+
+If the `-j`/`--junit` option is specified, `markdownlint-cli` will output an XML report in JUnit format.
+This report can be used in CI/CD environments to make the `markdownlint-cli` results available to your CI/CD tool.
+If both `-j`/`--junit` and `-o`/`--output` are specified, `-o`/`--output` will take precedence.
+Only one of these options should be specified.
 
 ## Configuration
 
@@ -100,6 +108,7 @@ A JS configuration file may internally `require` one or more npm packages as a w
 - `1`: Linting errors / bad parameter
 - `2`: Unable to write `-o`/`--output` output file
 - `3`: Unable to load `-r`/`--rules` custom rule
+- `4`: Unable to write `-j`/`--junit` output file
 
 ## Related
 
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -163,12 +163,15 @@ function printResult(lintResult) {
 
   if (program.junit) {
     const builder = require('junit-report-builder');
-    const suite = builder.testSuite().name('Markdownlint');
+    const reportStart = new Date();
+    const suite = builder.testSuite().name('Markdownlint').timestamp(reportStart.toISOString());
     results.forEach(result => {
       const {file, lineNumber, column, names, description} = result;
       const columnText = column ? `:${column}` : '';
-      suite.testCase().className(file).name(`${file}:${lineNumber}${columnText} ${names}`).failure(`${file}:${lineNumber}${columnText} ${names} ${description}`);
+      suite.testCase().className(file).name(`${file}:${lineNumber}${columnText} ${names}`).failure(`${file}:${lineNumber}${columnText} ${names} ${description}`, names).time(0.001);
     });
+    const reportTime = (new Date().getTime() - reportStart.getTime()) / 1000;
+    suite.time(reportTime);
     try {
       builder.writeTo(program.junit);
     } catch (error) {
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -164,11 +164,19 @@ function printResult(lintResult) {
   if (program.junit) {
     const builder = require('junit-report-builder');
     const reportStart = new Date();
-    const suite = builder.testSuite().name('Markdownlint').timestamp(reportStart.toISOString());
+    const suite = builder
+      .testSuite()
+      .name('Markdownlint')
+      .timestamp(reportStart.toISOString());
     results.forEach(result => {
       const {file, lineNumber, column, names, description} = result;
       const columnText = column ? `:${column}` : '';
-      suite.testCase().className(file).name(`${file}:${lineNumber}${columnText} ${names}`).failure(`${file}:${lineNumber}${columnText} ${names} ${description}`, names).time(0.001);
+      suite
+        .testCase()
+        .className(file)
+        .name(`${file}:${lineNumber}${columnText} ${names}`)
+        .failure(`${file}:${lineNumber}${columnText} ${names} ${description}`, names)
+        .time(0.001);
     });
     const reportTime = (new Date().getTime() - reportStart.getTime()) / 1000;
     suite.time(reportTime);
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -161,7 +161,23 @@ function printResult(lintResult) {
     process.exitCode = 1;
   }
 
-  if (program.output) {
+  if (program.junit) {
+    const builder = require('junit-report-builder');
+    const suite = builder.testSuite().name('Markdownlint');
+    results.forEach(result => {
+      const {file, lineNumber, column, names, description} = result;
+      const columnText = column ? `:${column}` : '';
+      suite.testCase().className(file).name(`${file}:${lineNumber}${columnText} ${names}`).failure(`${file}:${lineNumber}${columnText} ${names} ${description}`);
+    });
+    try {
+      builder.writeTo(program.junit);
+    } catch (error) {
+      console.warn('Cannot write to output file ' + program.junit + ': ' + error.message);
+      process.exitCode = 2;
+    }
+
+    console.log(lintResultString);
+  } else if (program.output) {
     try {
       fs.writeFileSync(program.output, lintResultString);
     } catch (error) {
@@ -185,6 +201,7 @@ program
   .option('-f, --fix', 'fix basic errors (does not work with STDIN)')
   .option('-s, --stdin', 'read from STDIN (does not work with files)')
   .option('-o, --output [outputFile]', 'write issues to file (no console)')
+  .option('-j, --junit [junitFile]', 'write issues to file in junit format as well as to the console')
   .option('-c, --config [configFile]', 'configuration file (JSON, JSONC, JS, or YAML)')
   .option('-i, --ignore [file|directory|glob]', 'file(s) to ignore/exclude', concatArray, [])
   .option('-p, --ignore-path [file]', 'path to file with ignore pattern(s)')
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -167,16 +167,21 @@ function printResult(lintResult) {
       .testSuite()
       .name('Markdownlint')
       .timestamp(new Date().toISOString());
-    results.forEach(result => {
-      const {file, lineNumber, column, names, description} = result;
-      const columnText = column ? `:${column}` : '';
-      suite
-        .testCase()
-        .className(file)
-        .name(`${file}:${lineNumber}${columnText} ${names}`)
-        .failure(`${file}:${lineNumber}${columnText} ${names} ${description}`, names)
-        .time(0);
-    });
+    if (results.length > 0) {
+      results.forEach(result => {
+        const {file, lineNumber, column, names, description} = result;
+        const columnText = column ? `:${column}` : '';
+        suite
+          .testCase()
+          .className(file)
+          .name(`${file}:${lineNumber}${columnText} ${names}`)
+          .failure(`${file}:${lineNumber}${columnText} ${names} ${description}`, names)
+          .time(0);
+      });
+    } else {
+      suite.testCase().className(program.args).name('Markdownlint');
+    }
+
     suite.time(0);
     try {
       builder.writeTo(program.junit);
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -163,11 +163,10 @@ function printResult(lintResult) {
 
   if (program.junit) {
     const builder = require('junit-report-builder');
-    const reportStart = new Date();
     const suite = builder
       .testSuite()
       .name('Markdownlint')
-      .timestamp(reportStart.toISOString());
+      .timestamp(new Date().toISOString());
     results.forEach(result => {
       const {file, lineNumber, column, names, description} = result;
       const columnText = column ? `:${column}` : '';
@@ -176,10 +175,9 @@ function printResult(lintResult) {
         .className(file)
         .name(`${file}:${lineNumber}${columnText} ${names}`)
         .failure(`${file}:${lineNumber}${columnText} ${names} ${description}`, names)
-        .time(0.001);
+        .time(0);
     });
-    const reportTime = (new Date().getTime() - reportStart.getTime()) / 1000;
-    suite.time(reportTime);
+    suite.time(0);
     try {
       builder.writeTo(program.junit);
     } catch (error) {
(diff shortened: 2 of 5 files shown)


DavidAnson:
Great, thanks, I’ll try to review this tonight! One thing you might do if bored is look at how other projects, especially ESLint, handle some of these things (like time stamp) in their jUnit output. Probably can’t go wrong being consistent. :)

FISHMANPET:
Also it turns out this will still throw an exit code of 1 when there are test failures because of https://github.com/igorshubovych/markdownlint-cli/blob/e857b964a04c8d81f660f65951573d52294fcc82/markdownlint.js#L161
I thought about wrapping that in something like `if (!program.junit)` or adding a new parameter like `--no-non-zero-exit-code-on-linting-errors` but you said you wanted to consider that separately.  I could open a new issue if you'd like to discuss that part.

FISHMANPET:
Good idea to look at what [eslint does](https://github.com/eslint/eslint/blob/master/lib/cli-engine/formatters/junit.js). I'll need to grok what they're doing a little bit, but at the very least it makes me realize that I'm not sure what will happen if there are no markdown errors, I think I need to add an explicit case like eslint to report a single successful test case if there are no issues.

DavidAnson:
Sorry, didn’t get to this again. Fully expect to tomorrow, thanks for your patience!

markdownlint.js
```diff
@@ -161,7 +161,37 @@ function printResult(lintResult) {
     process.exitCode = 1;
   }
 
-  if (program.output) {
```
DavidAnson, line 164: Can you leave the output section first as it is more common?

markdownlint.js
```diff
@@ -161,7 +161,37 @@ function printResult(lintResult) {
...
+  if (program.junit) {
+    const builder = require('junit-report-builder');
+    const suite = builder
+      .testSuite()
+      .name('Markdownlint')
```
DavidAnson, line 168: Lower case ‘m’, please.

markdownlint.js
```diff
@@ -161,7 +161,37 @@ function printResult(lintResult) {
     process.exitCode = 1;
   }
 
-  if (program.output) {
+  if (program.junit) {
+    const builder = require('junit-report-builder');
+    const suite = builder
```
DavidAnson, line 166: Mind calling it testSuite?

markdownlint.js
```diff
@@ -161,7 +161,37 @@ function printResult(lintResult) {
...
+        const columnText = column ? `:${column}` : '';
+        suite
+          .testCase()
+          .className(file)
+          .name(`${file}:${lineNumber}${columnText} ${names}`)
+          .failure(`${file}:${lineNumber}${columnText} ${names} ${description}`, names)
+          .time(0);
+      });
+    } else {
+      suite.testCase().className(program.args).name('Markdownlint');
```
DavidAnson, line 182: ‘m’ again

DavidAnson, line 182: Why program.args here?

markdownlint.js
```diff
@@ -161,7 +161,37 @@ function printResult(lintResult) {
...
+      .name('Markdownlint')
+      .timestamp(new Date().toISOString());
+    if (results.length > 0) {
+      results.forEach(result => {
+        const {file, lineNumber, column, names, description} = result;
+        const columnText = column ? `:${column}` : '';
+        suite
+          .testCase()
+          .className(file)
+          .name(`${file}:${lineNumber}${columnText} ${names}`)
```
DavidAnson, line 177: You could capture this in a variable and reuse it below.

markdownlint.js
```diff
@@ -161,7 +161,37 @@ function printResult(lintResult) {
...
+          .failure(`${file}:${lineNumber}${columnText} ${names} ${description}`, names)
+          .time(0);
+      });
+    } else {
+      suite.testCase().className(program.args).name('Markdownlint');
+    }
+
+    suite.time(0);
```
DavidAnson, line 185: Can you help me understand the difference between timestamp above and time(0) here and above?

markdownlint.js
```diff
@@ -161,7 +161,37 @@ function printResult(lintResult) {
...
+    try {
+      builder.writeTo(program.junit);
+    } catch (error) {
+      console.warn('Cannot write to output file ' + program.junit + ': ' + error.message);
```
DavidAnson, line 189: You could use the backtick string form here if desired.

markdownlint.js
```diff
@@ -161,7 +161,37 @@ function printResult(lintResult) {
...
+      process.exitCode = 2;
```
DavidAnson, line 190: Maybe update the docs or use a new exit code? https://github.com/igorshubovych/markdownlint-cli#exit-codes

markdownlint.js
```diff
@@ -161,7 +161,37 @@ function printResult(lintResult) {
...
+    }
+
+    console.log(lintResultString);
```
DavidAnson, line 193: Why not use console.error here as with the normal path?

markdownlint.js
```diff
@@ -182,9 +212,11 @@ program
   .version(pkg.version)
   .description(pkg.description)
   .usage('[options] <files|directories|globs>')
```
DavidAnson, line 214: Is .usage still needed after you add .arguments?

markdownlint.js
```diff
@@ -182,9 +212,11 @@ program
...
+  .arguments('[options] <files|directories|globs>')
   .option('-f, --fix', 'fix basic errors (does not work with STDIN)')
   .option('-s, --stdin', 'read from STDIN (does not work with files)')
-  .option('-o, --output [outputFile]', 'write issues to file (no console)')
+  .option('-o, --output <outputFile>', 'write issues to file (no console)')
```
DavidAnson, line 218: Thanks for fixing the brackets to show required value - mind doing the other options, too?

markdownlint.js
```diff
@@ -182,9 +212,11 @@ program
...
+  .option('-j, --junit [junitFile]', 'write issues to file in junit format as well as to the console')
```
DavidAnson, line 219: Is “junit” the preferred capitalization?

DavidAnson, line 219: I’d maybe drop the part “as well as ...”

package.json
```diff
@@ -44,6 +44,7 @@
     "ignore": "~5.1.4",
     "js-yaml": "~3.13.1",
     "jsonc-parser": "~2.2.0",
+    "junit-report-builder": "^2.0.0",
```
DavidAnson, line 47: ‘~’, please

markdownlint.js
DavidAnson, line 214: The relevant part of the docs should be updated: https://github.com/igorshubovych/markdownlint-cli#usage

markdownlint.js
FISHMANPET: My understanding is that program.args will be "everything else" in the command line that isn't part of an option, so in the case of markdownlint-cli it should be what's specified to be linted.  The className argument seems to generally be the "file" that the test was run against, and in the absence of errors I thought the files being linted would make sense as the "class name" but I'm open other suggestions here as well.

markdownlint.js
FISHMANPET: sure, makes sense

markdownlint.js
FISHMANPET: I figured there'd be little stylistic things here I wasn't aware of, changed

markdownlint.js
FISHMANPET: done

markdownlint.js
FISHMANPET: changed

markdownlint.js
FISHMANPET: looks like `JUnit` is the correct name, but I'm thinking it makes sense to leave the option name and file name in the current case, and just change the descriptive text?

FISHMANPET: removed

package.json
FISHMANPET: This is what I get for just blindly copy/pasting from the internet, updated 😄

markdownlint.js
FISHMANPET: Sure. The JUnit format has a timestamp indicating when the Test Suite was started, and then Time indicates the time (in seconds) that the test suite took.  Test cases also have a time in second, but no timestamp to indicate when they started.  The fact that `testSuite.time(0)` is here is I think a relic of when I was trying to actually calculate the time it took to generate the report, so I couldn't set the time until the report was done.  However after looking at what eslint did (just setting the time to 0) I removed the calculation but left setting it there.  There's no reason why it can't be moved into the construction of the testSuite since I know the time will be 0.

But regardless of where it's defined, they're two separate values.  Eslint just doesn't set the timestamp at all, but it was one of the JUnit fields, separate from time.

markdownlint.js
FISHMANPET: makes sense

markdownlint.js
FISHMANPET: I was thinking this was how the exit code was getting set on error but turns out that's not the case, so I'll switch it to error.

markdownlint.js
FISHMANPET: Nope, that was me playing around with commander and forgot to take that out.  It looks like commander doesn't have a way to define that "everything else" is a specific parameter so basically none of the parameter validation (which I think for this just means required vs optional) actually does anything.  And regardless it looks like `.arguments()` is only for commands so it wasn't doing anything here.  I'll remove it.

markdownlint.js
FISHMANPET: fixed

markdownlint.js
FISHMANPET: Set it to 4 and added to the readme as well.

markdownlint.js
FISHMANPET: Updated all of them

markdownlint.js
FISHMANPET: Added the new option, updated the required/optional brackets, and added a short section to the README explaining the JUnit output.

markdownlint.js
FISHMANPET: I just copied what was being done with the `output` option already, but it looks like in every case _except_ the console messages you're using backtick strings, so I replaced the 4 `+` constructed console messages with backtick strings for overall consistency.

FISHMANPET:
I pushed changes for all the comments I resolved in c4e3f57.  Looking at how tests are currently being done it shouldn't be too difficult to put some tests in.

Case/Situations I think should be tested:
JUnit report when there are issues (will pick one of your "incorrect" md test documents)
JUnit report when there aren't issues (will pick one of your "correct" md test documents)
When both `-o` and `-j` are specified, ensure that the `-o` section is what's run and the JUnit stuff will be skipped.

Anything else you'd like to see tested with these changes?

FISHMANPET:
OK, played around with this.  Decided to replicate the four `--output` tests with `--junit` and started with the first one, which was nothing in via stdin.  I settled on using an XML parser to parse the output XML, settling with xml-js (added as a dev dependency to package.json).  I figured I'd push the first test to see if you agree with the methodology before I did more.  I'm not sure if subsequent tests would be quite so rigorous, as it is that test checked everything that the builder writes, but subsequent tests should only need to test the things that are different from the first test.

Also writing the tests pointed out a few flaws (I love when that happens!) that I corrected in `markdownlint.js`.  Now if the input is stdin then the classname will be `stdin` instead of `program.args`.  And writing an empty string to console.error is different from not writing to console.error, so I set it to only write to the console if there are actual results.

DavidAnson:
I haven’t had a chance to look at any of this yet, but saw this latest message fly by and wanted to chime in in case it could save you some time. What do you think about just doing a plain text comparison of the output instead of trying to re-parse it as XML? I don’t imagine the layout or format will change unexpectedly, so does this give us enough confidence (assuming that we make sure the text output is correct)? Or, if you’ve already done this, maybe it’s best to be general and I do like that there will be guaranteed validation of the XML structure with a parser.

markdownlint.js
DavidAnson: Makes sense!

markdownlint.js
DavidAnson: Agreed!

README.md
```diff
@@ -63,6 +64,13 @@ Because this option makes changes to the input files, it is good to make a backu
 
 > Because not all rules include fix information when reporting errors, fixes may overlap, and not all errors are fixable, `--fix` will not usually address all errors.
 
+### JUnit report
+
+If the `-j`/`--junit` option is specified, `markdownlint-cli` will output an XML report in JUnit format.
```
DavidAnson, line 69: Add link to some info about “JUnit format”, maybe?

README.md
```diff
@@ -63,6 +64,13 @@ Because this option makes changes to the input files, it is good to make a backu
...
+This report can be used in CI/CD environments to make the `markdownlint-cli` results available to your CI/CD tool.
+If both `-j`/`--junit` and `-o`/`--output` are specified, `-o`/`--output` will take precedence.
+Only one of these options should be specified.
```
DavidAnson, line 72: Maybe remove this, I think your previous line is clear.

markdownlint.js
```diff
@@ -67,7 +67,7 @@ function readConfiguration(args) {
         markdownlint.readConfigSync(userConfigFile, configFileParsers);
       config = require('deep-extend')(config, userConfig);
     } catch (error) {
-      console.warn('Cannot read or parse config file ' + userConfigFile + ': ' + error.message);
+      console.warn(`Cannot read or parse config file ${userConfigFile}: ${error.message}`);
```
DavidAnson, line 70: Thanks for doing the others!

markdownlint.js
```diff
@@ -165,9 +165,43 @@ function printResult(lintResult) {
...
+        testSuite
+          .testCase()
+          .className(file)
+          .name(testName)
+          .failure(`${testName} ${description}`, names)
+          .time(0);
+      });
+    } else {
+      const className = program.stdin ? 'stdin' : program.args;
+      testSuite.testCase().className(className).name('markdownlint').time(0);
```
DavidAnson, line 192: Maybe make this multi-line for consistency with above?

markdownlint.js
```diff
@@ -184,11 +218,12 @@ program
...
-  .option('-c, --config [configFile]', 'configuration file (JSON, JSONC, JS, or YAML)')
-  .option('-i, --ignore [file|directory|glob]', 'file(s) to ignore/exclude', concatArray, [])
-  .option('-p, --ignore-path [file]', 'path to file with ignore pattern(s)')
-  .option('-r, --rules  [file|directory|glob|package]', 'custom rule files', concatArray, []);
+  .option('-o, --output <outputFile>', 'write issues to file (no console)')
+  .option('-j, --junit <junitFile>', 'write issues to file in JUnit format and to the console')
+  .option('-c, --config <configFile>', 'configuration file (JSON, JSONC, JS, or YAML)')
+  .option('-i, --ignore <file|directory|glob>', 'file(s) to ignore/exclude', concatArray, [])
+  .option('-p, --ignore-path <file>', 'path to file with ignore pattern(s)')
+  .option('-r, --rules  <file|directory|glob|package>', 'custom rule files', concatArray, []);
```
DavidAnson, line 226: Just noticed the extra space here, mind removing it?

test/test.js
```diff
@@ -357,6 +358,85 @@ test('--output with invalid path fails', async t => {
...
+  const input = '';
+  const junit = '../junitA.xml';
+  const result = await execa('../markdownlint.js',
+    ['--stdin', '--junit', junit],
+    {input, stripFinalNewline: false});
+  const xml = fs.readFileSync(junit, 'utf8');
+  const parsedXml = xmlParser.xml2js(xml, {compact: true});
+  t.is(result.stdout, '');
+  t.is(result.stderr, '');
+  t.is(Object.keys(parsedXml.testsuites).length, 1);
```
DavidAnson, line 371: See my separate comment about maybe just comparing the output text directly. What you have here is more correct and thorough, but it’s a lot more typing and harder to follow. I’m fine with either choice - up to you if you want to type more. :)

test/test.js
```diff
@@ -357,6 +358,85 @@ test('--output with invalid path fails', async t => {
...
+  t.is(parsedXml.testsuites.testsuite._attributes.failures, '0');
+  t.is(parsedXml.testsuites.testsuite._attributes.errors, '0');
+  t.is(parsedXml.testsuites.testsuite._attributes.skipped, '0');
+  t.is(parsedXml.testsuites.testsuite.testcase._attributes.classname, 'stdin');
+  t.is(parsedXml.testsuites.testsuite.testcase._attributes.name, 'markdownlint');
+  t.is(parsedXml.testsuites.testsuite.testcase._attributes.time, '0');
+  fs.unlinkSync(junit);
+});
+
+// WIP
```
DavidAnson, line 386: Your test cases make sense to me!

FISHMANPET:
As is the case with open source, I got distracted from this for a while and now (exactly a year later!) I'm looking at it again.

I believe at the time I got caught up on tests, and decided to implement an overly rigorous test regime (using the XML parser).  You recommended just comparing to a known file and... I'm not sure why I ignored that at the time.

I also see you've adapted much of this into markdownlint-cli2 so I'm not sure there's any benefit to "finishing" this.  If you'd like I could redo the tests to just match against a known file, and align the output with what you've done with the junit formatter in cli2. But I read your post about why you wrote cli2, and my guess is you'd be fine just closing this without merging, since the meat of the work has been implemented there.

DavidAnson:
Glad to have you back! What do you do here is kind of up to you! I wrote CLI2 for the reasons you read about, but I don't expect everybody to migrate over immediately - or even ever. Having a compatible implementation for this project would be nice and probably benefit some people. On the other hand, not having it represents a small carrot for people to migrate. I'm happy to support whatever approach you choose. :)

DavidAnson:
Closing per discussion above.

PR #114, opened 2020-11-05
README: add Homebrew installation instructions

### LINKED ISSUE #111: Consider adding homebrew installation option
When developing a project that has no package.json (for example, a shell script project), it would be nice to be able to include `markdownlint` in the project's brew bundle file instead of having to create a package.json just for it. This is something [Prettier](https://github.com/prettier/prettier), a project similar to yours, is doing. See [here](https://github.com/Homebrew/homebrew-core/blob/e9e4aa606c048454ef19ce16c2e71652984c9e73/Formula/prettier.rb).

Apart from writing the initial fomula (probably very similar to the one I linked), you'd have to include a way to update the hashes when you release a new version. I don't know what are the best practices for automating that, but my guess would be to do it through a github action.

Would you be open to add this feature to the project? If so, would you accept pull requests for it?

dtrodrigues:
Adds installation instructions for Homebrew: https://formulae.brew.sh/formula/markdownlint-cli

Resolves #111

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,12 @@
 npm install -g markdownlint-cli
 ```
 
+On macOS you can install via [Homebrew](https://brew.sh/):
+
+```bash
+brew install markdownlint-cli
+```
+
 ## Usage
 
 ```bash


DavidAnson:
Thanks!

Could you maybe update the message not to say "latest version"? I don't think anyone is signed up to update brew with every release.

Also, why is brew using Node 15 (latest) instead of Node 14 (recommended)?

dtrodrigues:
I removed the "latest version" wording, but usually a Homebrew contributor will submit a PR to update to the laetest version relatively quickly after a new release. It doesn't have to, and in general does not, come from the upstream repository members itself.

Brew always tries to use the latest stable software. If markdownlint is known to not work with Node 15, it can be pinned to Node 14. I've run it locally against Node 15 and it seems to work okay except one test doesn't pass since the error message slightly changed: https://travis-ci.com/github/dtrodrigues/markdownlint-cli/jobs/428773281

Node 14:
```
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! markdownlint-cli@0.24.0 invalid: `node ./markdownlint.js --config test/test-config.json -- test/incorrect.md`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the markdownlint-cli@0.24.0 invalid script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

Node 15:
```
npm ERR! code 1
npm ERR! path /Users/xxx/markdownlint-cli
npm ERR! command failed
npm ERR! command sh -c node ./markdownlint.js --config test/test-config.json -- test/incorrect.md
```

and https://github.com/igorshubovych/markdownlint-cli/blob/b0fe91bb57e3872754d0c46323d859f413909d1e/test/test.js#L89 is looking for `npm ERR! code ELIFECYCLE` in the `npm run invalid` output.

nschonni:
Test failure on 15 is likely related to NPM changing from 6 to 7

DavidAnson:
Thank you! Regarding Node versions, my point was that v15 is NOT the latest stable version. But if that's what Brew prefers, it should be fine. I'll update the tests for Node 15 soon.

PR #134, opened 2020-11-24
feat: only fix specific rules

dsanders11:
I have a use-case where I'd only like to automatically fix specific rules (just ones related to whitespace really) without necessarily fixing all rules.

This PR makes that possible with a new `--fix-rule` flag, which can be used multiple times to specify multiple rules.

I don't really love the flag name and all, but was trying to match the style of the other flags, like `--rules`. I'd personally probably prefer a comma-separated list of rules since it's less verbose, maybe `--fix-rules [rule,]`, or just make `--fix` take an optional list of rules, otherwise fix all.

So, feedback welcome on the flag.

Can update `README.md` and add a test into this PR once feedback solidifies the final form for the flag.

### DIFF
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -183,6 +183,7 @@ program
   .description(pkg.description)
   .usage('[options] <files|directories|globs>')
   .option('-f, --fix', 'fix basic errors (does not work with STDIN)')
+  .option('--fix-rule [rule]', 'only fix a specific rule (does not work with STDIN)', concatArray, [])
   .option('-s, --stdin', 'read from STDIN (does not work with files)')
   .option('-o, --output [outputFile]', 'write issues to file (no console)')
   .option('-c, --config [configFile]', 'configuration file (JSON, JSONC, JS, or YAML)')
@@ -272,7 +273,7 @@ function lintAndPrint(stdin, files) {
     };
   }
 
-  if (program.fix) {
+  if (program.fix || program.fixRule) {
     const fixOptions = {
       ...lintOptions,
       resultVersion: 3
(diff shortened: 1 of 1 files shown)


dsanders11:
@DavidAnson, any thoughts on moving this forward? I'd like to setup `markdownlint` for Electron's docs, and this functionality would be useful for that change being a drop-in replacement.

DavidAnson:
The pull request appears to fail most of the unit tests, so I haven't looked at it.

Also, I think you can accomplish what do you want today by disabling rules you do not want to fix and using the existing `fix` parameter. You could keep a dedicated configuration file for this purpose.

dsanders11:
> The pull request appears to fail most of the unit tests, so I haven't looked at it.

I noticed that, but wasn't going to look into it until some discussion happened since there's open-ended questions, and the implementation may change.

> Also, I think you can accomplish what do you want today by disabling rules you do not want to fix and using the existing fix parameter. You could keep a dedicated configuration file for this purpose.

You *could* do that I suppose, but is that a reason not to have this functionality? That's a very clunky way to do it, and fragile if new rules are added to `markdownlint` since it wouldn't be clear that there are new rules now enabled which need to be disabled in the separate configuration.

DavidAnson:
For my part, I would generally rather not support multiple ways to do a thing because that's more code to write, test, and support. In this case, it seems straightforward, clear, and reliable to do what you want using the approach I suggest above:

```bash
pi@pi:~/issue134 $ cat sample.md 
# Sample

Line with extra whitespace 

Line with <html/> element
pi@pi:~/issue134 $ ./node_modules/.bin/markdownlint sample.md 
sample.md:3:27 MD009/no-trailing-spaces Trailing spaces [Expected: 0 or 2; Actual: 1]
sample.md:5:11 MD033/no-inline-html Inline HTML [Element: html]
pi@pi:~/issue134 $ cat .markdownlint-fix.json 
{
  "default": false,
  "whitespace": true
}
pi@pi:~/issue134 $ ./node_modules/.bin/markdownlint --config .markdownlint-fix.json --fix sample.md 
pi@pi:~/issue134 $ cat sample.md 
# Sample

Line with extra whitespace

Line with <html/> element
pi@pi:~/issue134 $ ./node_modules/.bin/markdownlint sample.md 
sample.md:5:11 MD033/no-inline-html Inline HTML [Element: html]
pi@pi:~/issue134 $ 
```

dsanders11:
Thanks for the example. I wasn't aware of the `default` property for the config, that handles my concern about new rules coming online unexpectedly, and keeps the config small and concise. I'll go ahead and close this since it's not functionality you'd like to implement.

DavidAnson:
Cool, thanks!

PR #153, opened 2021-01-06
Add support for fixing on stdin

technicalpickles:
This works by reading from stdin, and then applying fixes, and printing to the stdout. It will output the fixed file if --output is given.

The motivation for this is to be able to use it as a pipe, ie:

```terminal
$ pbpaste | markdownlint --fix --stdin | pbcopy
```

### DIFF
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -128,8 +128,8 @@ function prepareFileList(files, fileExtensions, previousResults) {
   });
 }
 
-function printResult(lintResult) {
-  const results = flatten(Object.keys(lintResult).map(file => {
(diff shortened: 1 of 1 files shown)


DavidAnson:
Thanks for the proposal! I've only looked briefly on my phone just now. One thing is that we'll need tests for this new code. Another is that it's not clear to me that results are output to STDERR in the new flow?

technicalpickles:
> Another is that it's not clear to me that results are output to STDERR in the new flow?

It does not, but that is something I was thinking 👍🏻

PR #155, opened 2021-01-10
Add End of File Newline to program.output

Kangie:
The output file specified with the `-o` or `--output` flags should contain an EOF newline to be a valid POSIX file.

This PR adds an EOF newline to the output.

### DIFF
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -163,19 +163,17 @@ function printResult(lintResult) {
   }
 
   if (program.output) {
-      try {
-        if (lintResultString.length > 0) {
-          fs.writeFileSync(program.output, lintResultString + os.EOL); // We need to append a newline to make a valid POSIX file.
-        }
-        else {
-          fs.writeFileSync(program.output, lintResultString); // This is effectively writing a blank file.
-        }
-      } catch (error) {
-        console.warn('Cannot write to output file ' + program.output + ': ' + error.message);
-        process.exitCode = 2;
-      }
-  } else if (lintResultString) {
-    console.error(lintResultString);
+    lintResultString = lintResultString.length > 0 ?
+      lintResultString + os.EOL :
+      lintResultString
+    try {
+      fs.writeFileSync(program.output, lintResultString);
+    } catch (error) {
+      console.warn('Cannot write to output file ' + program.output + ': ' + error.message);
+      process.exitCode = 2;
+    }
+    } else if (lintResultString) {
+      console.error(lintResultString);
   }
 }
 
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -14,6 +14,7 @@ const glob = require('glob');
 const minimatch = require('minimatch');
 const minimist = require('minimist');
 const pkg = require('./package');
+const os = require ('os');
 
 function jsoncParse(text) {
   return JSON.parse(require('jsonc-parser').stripComments(text));
@@ -162,22 +163,17 @@ function printResult(lintResult) {
   }
 
   if (program.output) {
-    if (lintResultString.length > 0) {
       try {
-        fs.writeFileSync(program.output, lintResultString + '\n'); // We need to append a newline to make a valid POSIX file.
+        if (lintResultString.length > 0) {
+          fs.writeFileSync(program.output, lintResultString + os.EOL); // We need to append a newline to make a valid POSIX file.
+        }
+        else {
+          fs.writeFileSync(program.output, lintResultString); // This is effectively writing a blank file.
+        }
       } catch (error) {
         console.warn('Cannot write to output file ' + program.output + ': ' + error.message);
         process.exitCode = 2;
       }
-    }
-    else {
-      try {
-        fs.writeFileSync(program.output, lintResultString); // This is effectively writing a blank file.
-      } catch (error) {
-        console.warn('Cannot write to output file ' + program.output + ': ' + error.message);
-        process.exitCode = 2;
-      }
-    }
   } else if (lintResultString) {
     console.error(lintResultString);
   }
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -172,7 +172,7 @@ function printResult(lintResult) {
       console.warn('Cannot write to output file ' + program.output + ': ' + error.message);
       process.exitCode = 2;
     }
-    } else if (lintResultString) {
+  } else if (lintResultString) {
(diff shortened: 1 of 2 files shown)


Kangie:
Obviously the tests do not like having a newline at the end of the output file.

I'll take a deeper look into it, but it appears that two things need to happen at the moment:

1) A special case should be added for "no output" that does not append a newline.
2) Any tests that expect the output file to have content should also expect the file to end with a newline.

Kangie:
That should take care of it. The newline is only appended if there's actually content now.

DavidAnson:
Could you please link to the specification that requires a trailing new line character? Also, if we move ahead with this, a test case to verify the change would be great. Thanks!

nschonni:
This seems to be what https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md047---files-should-end-with-a-single-newline-character is for

Kangie:
Hi @DavidAnson 

[POSIX Specification](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206)

The POSIX definition of a text file is: "A file that contains characters organized into one or more lines."

The POSIX definition of a line is: "A sequence of zero or more non- <newline> characters plus a terminating <newline> character."

Therefore, a valid POSIX text file must contain a terminating newline character as per this change. :)'

I'll write a test for "Invalid markdown input must result in a file that ends in \n" and add it to the PR.

DavidAnson:
I'd not seen that, thanks! I'll post a review once you add the test. FYI, I'll ask about reducing some of the redundancy in the current diff.

Kangie:
OK, I'm still not able to run tests locally (I think that the ava binary isn't in my path) so there may be a few commits while I get the syntax down.

Kangie:
OK. I've amended the way that this operates to use the node 'os' lib's 'EOL' function to pick the appropriate EOL based on the platform in use. This should work better across operating systems!

I can see that tests are passing now, so I'll rebase to tidy up the commits and it should be good to review!

Kangie:
I didn't review the package-lock.json update - It's huge, I suspect because it's going to v2 - It's just the output from npm but I'm happy to remove from the PR if you'd prefer to do that yourself / have it as a separate PR.

test/test.js
```diff
@@ -725,3 +726,17 @@ test('--ignore-path fails for missing file', async t => {
     t.regex(error.stderr, /enoent.*no such file or directory/i);
   }
 });
+
+test('Linter text file --output must end with EOF newline', async t => {
+  const output = '../outputE.txt';
```
DavidAnson, line 731: Thanks for adding this and using a unique name!

test/test.js
```diff
@@ -725,3 +726,17 @@ test('--ignore-path fails for missing file', async t => {
...
+  let endOfLine = new RegExp(os.EOL + '$');
```
DavidAnson, line 732: const

test/test.js
```diff
@@ -725,3 +726,17 @@ test('--ignore-path fails for missing file', async t => {
...
+  try {
+    await execa('../markdownlint.js',
+      ['--config', 'test-config.json', '--output', output, 'incorrect.md'],
+      {stripFinalNewline: false});
+    t.fail();
+  } catch (error) {
+    t.is(endOfLine.test(fs.readFileSync(output, 'utf8')), true);
```
DavidAnson, line 739: `t.true(check())` rather than `t.is(check(), true)`

markdownlint.js
```diff
@@ -162,12 +163,17 @@ function printResult(lintResult) {
...
   if (program.output) {
-    try {
-      fs.writeFileSync(program.output, lintResultString);
-    } catch (error) {
-      console.warn('Cannot write to output file ' + program.output + ': ' + error.message);
-      process.exitCode = 2;
-    }
+      try {
+        if (lintResultString.length > 0) {
+          fs.writeFileSync(program.output, lintResultString + os.EOL); // We need to append a newline to make a valid POSIX file.
```
DavidAnson, line 168: Could this become a 1-line change with something like:
`fs.writeFileSync(program.output, lintResultString + (lintResultString.length > 0 ? os.EOL : '');`

markdownlint.js
```diff
@@ -162,12 +163,17 @@ function printResult(lintResult) {
   }
 
   if (program.output) {
-    try {
-      fs.writeFileSync(program.output, lintResultString);
-    } catch (error) {
-      console.warn('Cannot write to output file ' + program.output + ': ' + error.message);
-      process.exitCode = 2;
-    }
+      try {
```
DavidAnson, line 166: There looks to be an extra indent for this block of lines - combined with my other comment here, I think there's just a single line change?

DavidAnson (changes requested):
Thanks for the contribution! Just a couple of small comments.

DavidAnson:
Oh, and please exclude `package-lock.json` from the diff - that should not be changing here.

markdownlint.js
Kangie: The meaning is not as clear, and it's use is [contentious](https://eslint.org/docs/rules/no-ternary). Many maintainers disallow it, so I default to the longer form.

I've refactored the code to use a long-form ternary before we attempt to write out. It should be a bit better. Let me know what you think!

markdownlint.js
```diff
@@ -162,14 +163,17 @@ function printResult(lintResult) {
...
+      lintResultString
     try {
       fs.writeFileSync(program.output, lintResultString);
     } catch (error) {
       console.warn('Cannot write to output file ' + program.output + ': ' + error.message);
       process.exitCode = 2;
     }
-  } else if (lintResultString) {
-    console.error(lintResultString);
+    } else if (lintResultString) {
```
DavidAnson, line 175: Could you please undo this new indent?

markdownlint.js
```diff
@@ -162,14 +163,17 @@ function printResult(lintResult) {
   }
 
   if (program.output) {
+    lintResultString = lintResultString.length > 0 ?
```
DavidAnson, line 166: This works, thanks!

DavidAnson (changes requested):
One really minor issue left, if you could please.

Kangie:
@DavidAnson

Not sure how I messed up that indent, but it should be good to go now!

DavidAnson (approved):

PR #167, opened 2021-02-21
feat: add dot option to treat dots as normal characters

Divlo:
Hey! 👋
Thanks for this CLI tool to use `markdownlint`. :+1: 

I'm using the CLI tool like this : `markdownlint '**/*.md' --ignore node_modules` but currently folders like `.github` are ignored, so my `CONTRIBUTING.md`, `PULL_REQUEST_TEMPLATE.md` etc files are ignored.
The current workaround is to use the CLI like this : `markdownlint '**/*.md' '.*/**/*.md' --ignore node_modules`.

I don't think it is wanted, in my opinion, when we're setting this globbing `'**/*.md'`, it means we want to lint **EVERY** markdown files in the project, regardless of their location in the folder structure, including folders with a dot.

The fix seems simple, we need to set `dot: true` in the `glob` options.
See: <https://www.npmjs.com/package/glob#dots>

Just to test I edited myself the source files in the `node_modules` with the fix of this PR, and it is indeed fixing the issue.

### DIFF
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -79,7 +79,7 @@ function readConfiguration(args) {
 function prepareFileList(files, fileExtensions, previousResults) {
   const globOptions = {
     nodir: true,
-    dot: true
+    dot: options.dot != null ? true : false
   };
   let extensionGlobPart = '*.';
   if (fileExtensions.length === 1) {
@@ -194,7 +194,8 @@ program
   .option('-c, --config [configFile]', 'configuration file (JSON, JSONC, JS, or YAML)')
   .option('-i, --ignore [file|directory|glob]', 'file(s) to ignore/exclude', concatArray, [])
   .option('-p, --ignore-path [file]', 'path to file with ignore pattern(s)')
-  .option('-r, --rules  [file|directory|glob|package]', 'custom rule files', concatArray, []);
+  .option('-r, --rules  [file|directory|glob|package]', 'custom rule files', concatArray, [])
+  .option('-d, --dot', 'include files/folders with a dot (for example `.github`)');
 
 program.parse(process.argv);
 
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -78,8 +78,8 @@ function readConfiguration(args) {
 
 function prepareFileList(files, fileExtensions, previousResults) {
   const globOptions = {
-    nodir: true,
-    dot: options.dot != null ? true : false
+    dot: !!options.dot,
+    nodir: true
   };
   let extensionGlobPart = '*.';
   if (fileExtensions.length === 1) {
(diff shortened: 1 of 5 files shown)


DavidAnson:
This change makes sense, but I would like to hear from others before making it. The conventional behavior of dot files/directories is to be hidden by default and that's how the tool currently behaves. As you show above, it is already possible to override this behavior if that's what you want.

If we agree to change this behavior, one approach is what you've shown here and another would be to add a new option for this. The advantage of a new option would be that the default behavior does not change and remains consistent with dot file convention - rather than forcing everyone into this new behavior.

Divlo:
I agree, we should avoid breaking changes! :smile:
Your suggestion of adding a new option is great, so it doesn't force everyone with this new behavior. :+1: 
I just committed with a new boolean option with `commander` : `--dot`. @DavidAnson

markdownlint.js
```diff
@@ -78,7 +78,8 @@ function readConfiguration(args) {
 
 function prepareFileList(files, fileExtensions, previousResults) {
   const globOptions = {
-    nodir: true
+    nodir: true,
+    dot: options.dot != null ? true : false
```
Divlo, line 82: This line of code could be refactored like this `dot: options.dot ?? false` 
But you're supporting `node >=10`, and the nullish coalescing operator (??), only has been introduced in `node >= 14` according to [node.green](https://node.green/).

DavidAnson: I think `!!options.dot` is the most conventional way to do this.

DavidAnson: Also, please sort alphabetically - here and in the list of options in MD and JS.

DavidAnson (changes requested):
In addition to the minor comments above, there should be two new test cases to verify this behavior. One should verify a dot file and dot directory is ignored without the option and the other should verify the file and directory are included with the option. Thanks!

Divlo:
Sure, it's always better to add automated tests! :smile: 
I added the two new test cases, but I didn't quite understood what you mean by that :
> Also, please sort alphabetically - here and in the list of options in MD and JS.

The options weren't sorted alphabetically, I sorted all the options correctly, feel free to tell me if it's wrong or you thought of something else.

DavidAnson:
This looks good, thank you! As I mentioned above, I'd like to give other people a couple of days to have a look and see if they agree. In the meantime, it might be nice to add something like `.file-with-dot.md` for the new tests you added to confirm that it behaves correctly like the folder does.

Divlo:
Right, I added `.file-with-dot.md`. :+1: 
Thanks for your help! :smile: 

>  I'd like to give other people a couple of days to have a look and see if they agree

Now it's only a new option, and it doesn't affect the old behavior so we could merge this faster, but it's up to you of course! @DavidAnson

DavidAnson (approved):
Great! Yes, I feel better as an option, and will include this with the upcoming release.

PR #181, opened 2021-04-06
Switch rc dependency to run-con to address CVE-2020-7788 (fixes #146).

### LINKED ISSUE #146: `rc` is unhealthy dependency
[rc](https://www.npmjs.com/package/rc) package seems to be abandoned by it's author years ago and doesn't get the security updates : https://github.com/dominictarr/rc/issues

If it can't be brought back to life then it would be better to replace it with something else.

DavidAnson:
Submitting from phone, so untested.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3950,7 +3950,8 @@
     "ini": {
       "version": "1.3.7",
       "resolved": "https://registry.npmjs.org/ini/-/ini-1.3.7.tgz",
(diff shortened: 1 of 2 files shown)


DavidAnson:
Passing on Node 15 is a good sign. I can’t fix `package-lock.json` from my phone. Will do so later today and merge if everything passes.

PR #195, opened 2021-06-09
Add option to output results in JSON format

EliSchleifer:
1) Added '-j --json' option to stringify any detected issues as JSON string. 
2) Added tests to verify correct behavior of --json when outputting to file and outputting to stderr
3) Update README.md to list new option

Motivation:
Very common for linting frameworks to provide output format that is more machine readable. 

Thanks


DavidAnson:
My guess is that tests are failing because you did not use a unique name for the output file?

EliSchleifer:
Good call. I didn’t notice that the files weren’t temp/cleaned up.

Thanks David

________________________________

DavidAnson:
Could you please make sure the output exactly matches that of the corresponding CLI2 JSON formatter: https://www.npmjs.com/package/markdownlint-cli2-formatter-json

EliSchleifer:
Sure. Should be easy enough

________________________________

EliSchleifer:
Should be all set now. Much nicer with the full json information in there.

README.md
```diff
@@ -34,6 +34,7 @@ markdownlint --help
     -f, --fix                                   fix basic errors (does not work with STDIN)
     -i, --ignore [file|directory|glob]          file(s) to ignore/exclude
     -o, --output [outputFile]                   write issues to file (no console)
+    -j, --json                                  write issues in json format
```
DavidAnson, line 37: Please sort alphabetically.

test/test.js
```diff
@@ -70,6 +70,21 @@ test('linting of incorrect Markdown file fails', async t => {
...
+    await execa('../markdownlint.js',
+      ['--config', 'test-config.json', 'incorrect.md', '--json'],
+      {stripFinalNewline: false});
+    t.fail();
+  } catch (error) {
+    t.is(error.stdout, '');
+    const issues = JSON.parse(error.stderr);
+    t.is(issues.length, 8);
+    t.is(issues[0].fileName, "incorrect.md");
+
```
DavidAnson, line 84: Extra space

test/test.js
```diff
@@ -70,6 +70,21 @@ test('linting of incorrect Markdown file fails', async t => {
...
+  try {
+    await execa('../markdownlint.js',
+      ['--config', 'test-config.json', 'incorrect.md', '--json'],
+      {stripFinalNewline: false});
+    t.fail();
+  } catch (error) {
+    t.is(error.stdout, '');
+    const issues = JSON.parse(error.stderr);
+    t.is(issues.length, 8);
+    t.is(issues[0].fileName, "incorrect.md");
```
DavidAnson, line 83: Could you please validate the complete structure of one object so it won't change unexpectedly?

test/test.js
```diff
@@ -343,6 +358,27 @@ test('--output with invalid input outputs violations', async t => {
...
 });
 
+test('--output with invalid input and --json outputs issues as json', async t => {
+  const input = [
+    'Heading',
+    '',
+    'Text ',
+    ''
+  ].join('\n');
+  const output = '../outputC.json';
```
DavidAnson, line 368: Unique file name or this will fail randomly

markdownlint.js
```diff
@@ -278,6 +296,9 @@ function lintAndPrint(stdin, files) {
       stdin
     };
   }
+  if (options.json) {
+    lintOptions.resultVersion = 3;
```
DavidAnson, line 300: Can't easily review this file from my phone, will do so later...

DavidAnson (changes requested):
Looks great! Left some first draft comments from my phone just now.

README.md
EliSchleifer: sure. I had missed the fact they were sorted.

test/test.js
EliSchleifer: yup

test/test.js
EliSchleifer: This filename is unique - .json extension. But I changed it to outputJsonC.json to make it more explicit.

markdownlint.js
EliSchleifer: Cool thanks.

test/test.js
EliSchleifer: Sure.

README.md
DavidAnson: I think I'm looking at the latest changes, but it's still after -o instead of before it?

markdownlint.js
```diff
@@ -131,30 +131,47 @@ function prepareFileList(files, fileExtensions, previousResults) {
 }
 
 function printResult(lintResult) {
+  let lintResultString = '';
```
DavidAnson, line 134: It doesn't look like this needed to move? Not a problem, just curious why?

markdownlint.js
```diff
@@ -131,30 +131,47 @@ function prepareFileList(files, fileExtensions, previousResults) {
...
-      return a.file.localeCompare(b.file) || a.lineNumber - b.lineNumber ||
-        a.names.localeCompare(b.names) || a.description.localeCompare(b.description);
-    });
-    lintResultString = results.map(result => {
-      const {file, lineNumber, column, names, description} = result;
-      const columnText = column ? `:${column}` : '';
-      return `${file}:${lineNumber}${columnText} ${names} ${description}`;
-    }).join('\n');
+    if (options.json) {
+      results.sort((a, b) => {
```
DavidAnson, line 158: I wonder if we could reuse the same call to `sort` here? I think you're using `fileName` for consistency with CLI2? Then what about changing the existing `file` to `fileName` since it's internal only? Then if we did something like `((a.names && b.names) ? a.names.localeCompare(b.names) : 0)` and reuse the call to `sort`?

DavidAnson: Put differently: call `results.sort`, then `if (options.json) { stringify } else { map }`

README.md
DavidAnson: Oh, it's sorted in the code, but not the README here.

markdownlint.js
```diff
@@ -191,7 +208,8 @@ program
   .option('-d, --dot', 'include files/folders with a dot (for example `.github`)')
   .option('-f, --fix', 'fix basic errors (does not work with STDIN)')
   .option('-i, --ignore [file|directory|glob]', 'file(s) to ignore/exclude', concatArray, [])
-  .option('-o, --output [outputFile]', 'write issues to file (no console)')
+  .option('-j, --json', 'write issues in json format')
+  .option('-o, --output [outputFile]', 'write issues to file (no console)')
```
DavidAnson, line 212: There are 2 trailing spaces on this line that make it show in the diff.

test/test.js
```diff
@@ -70,6 +70,35 @@ test('linting of incorrect Markdown file fails', async t => {
...
+    const expected = {
+      fileName: "incorrect.md",
+      lineNumber: 1,
+      ruleNames: [
+        "MD002",
+        "first-heading-h1",
+        "first-header-h1",
+      ],
+      ruleDescription: "First heading should be a top-level heading",
+      ruleInformation: "https://github.com/DavidAnson/markdownlint/blob/v0.23.1/doc/Rules.md#md002",
```
DavidAnson, line 92: The version number here will break in the future - what I do in the library is put `v0.0.0` in the test and run a quick RegExp replace in the test before comparing.

DavidAnson (changes requested):
Thanks for making those changes! A few more comments and one refactoring idea to reuse code a little better in this pass. I don't expect anything more requests of significance.

markdownlint.js
EliSchleifer: Yeah - I wanted to write it that way, but the names and ruleNames are not the same structure form. One is collapsed with a join, the other is a list. So while fileName is totally reusable for the sort, the other tie-breakers - names, description are not stable. Maybe it doesn't matter. Happy to use just the fileName, lineNumber (past that point I am not sure it matters). 

Your project, your call.

markdownlint.js
EliSchleifer: Yeah I moved it in an earlier iteration when I tried to refactor a bit. No real value now to move it.

markdownlint.js
EliSchleifer: Got it

DavidAnson (approved):

PR #208, opened 2021-07-20
Add publish job to ci workflow

### LINKED ISSUE #205: Publish "official" container image for markdownlint-cli
Would there be any interest in publishing an "offical" docker image for markdownlint-cli? Right now I see a number of entries on dockerhub in personal repositories or the other option of building it yourself. It would be great to just rely on an already published image that has a higher level of trust (direct from the source!).

If there is interest, I have workflows from other projects that could just be tied into github actions and images could be published on merge (`latest`) and when a release is cut (`vX.Y.Z`). This could also all just be done directly in GitHub by publishing the image to the GH Container Registry if the overhead of a repo in docker or quay is too much.

Just wanted to gauge interest before opening a PR.

tylerauerbeck:
This PR adds a job to the ci workflow that builds an image using the latest build to a specified container registry and tags it with the latest tag. This job only runs when the ci job is successful and when a push occurs to the master branch (ensuring that all tests have passed).

The only requirements this has is that the container registry of choice is specified in the environment vars for the job and that two secrets are created in the repository (`REGISTRY_USERNAME` and `REGISTRY_PASSWORD`).

You can see the output of this workflow here (where I had specified my own repo/credentials):

https://quay.io/repository/tauerbec/markdownlint-cli?tab=tags
https://github.com/tylerauerbeck/markdownlint-cli/runs/3114894787?check_suite_focus=true

This also fixes a small issue I had encountered with the `.dockerignore` file and bumps the Dockerfile to build from `node:14-alpine`

Fixes #205


markdownlint.js
```diff
@@ -10,7 +10,7 @@ const options = program.opts();
 const differenceWith = require('lodash.differencewith');
 const flatten = require('lodash.flatten');
 const markdownlint = require('markdownlint');
-const rc = require('rc');
+const rc = require('run-con');
```
DavidAnson, line 13: This is no longer necessary

.github/workflows/ci.yml
```diff
@@ -36,3 +36,35 @@ jobs:
 
       - name: Run tests
         run: npm test
+
+  publish:
+    runs-on: ubuntu-latest
+    needs: test
+    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```
DavidAnson, line 43: Does this mean every push to master updates Docker? Shouldn't that be limited to releases of the package (i.e., to npm)?

.github/workflows/ci.yml
```diff
@@ -36,3 +36,35 @@ jobs:
...
+        uses: docker/setup-qemu-action@v1.2.0
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1.5.1
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1.10.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ secrets.REGISTRY_USERNAME }}
```
DavidAnson, line 62: Who provides these secrets and maintains them?

.github/workflows/ci.yml
```diff
@@ -36,3 +36,35 @@ jobs:
 
       - name: Run tests
         run: npm test
+
+  publish:
```
DavidAnson, line 40: publish-to-docker

.dockerignore
```diff
@@ -5,6 +5,6 @@ test/
 .editorconfig
 .travis.yml
 appveyor.yml
-node_modules\
-dist\
-node_modules
\ No newline at end of file
+node_modules/
+dist/
+node_modules
```
DavidAnson, line 10: Do we need this with and without the slash?

.github/workflows/ci.yml
DavidAnson: Or can this be in its own file?

DavidAnson (changes requested):
Thanks! I'm not familiar with this process and left a few questions.

.github/workflows/ci.yml
tylerauerbeck: So I went with each push to master so that if someone wants to use what is in the master branch, it's immediately available and it isn't put on the user to build this from scratch. Once this is in place, we could then add a release workflow that would also create a tagged version of the image.

.github/workflows/ci.yml
tylerauerbeck: This would be on the repo owner(s). They just need to be created and maintained by creating secrets in the repo settings.

.github/workflows/ci.yml
tylerauerbeck: It's better to have this in the same file and just use various conditionals to trigger this so that we make sure we're only building this when other CI passes.

.dockerignore
tylerauerbeck: I don't think so. But I was just looking to correct the error I was running into and didn't know if there was some history behind having them in there both ways.

.github/workflows/ci.yml
tylerauerbeck: Fixed the name as requested.

.dockerignore
DavidAnson: Good point, I didn't see this was there already

.github/workflows/ci.yml
DavidAnson: I think I'd prefer doing releases first - or at least together. I'd also prefer a trigger, I think - most of the commits are Dependabot and there's not much point publishing a new release because a devDependency updated.

.github/workflows/ci.yml
DavidAnson: I do not (currently) have Docker credentials.

tylerauerbeck: Gotcha. At least from what I was able to gather, I don't think there is an existing docker/quay/etc. repository for this. So it would be as simple as just creating one? But I may be wrong here.

DavidAnson (changes requested):
FYI, as a non-Docker user, I'm kind of tentative about this. It's me, not you.

.github/workflows/ci.yml
tylerauerbeck: That makes sense. I can make some changes to make it happen that way and then if we see that there is a need to make more frequent releases we can always add it back later.

DavidAnson:
@tylerauerbeck, after some Googling, it seems the following two pages outline how to use the GitHub container registry from an action:
- https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry
- https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions

I may need to tweak a permission on the repo, but otherwise this seems to enable the login, build, and publish of a container without explicitly managing or sharing accounts and passwords.

wburningham:
Any update on this? I'd love to use a docker image published from this repo.

DavidAnson:
If no one beats me to it, I will have a look when I get to this project in my rotation. Probably a couple of weeks or so.

tylerauerbeck:
@DavidAnson Sorry, this one fell off my radar. I'm happy to update this to publish to GHCR if you'll have some time for review.

DavidAnson:
It's a deal! :)

DavidAnson:
As of `v0.31.1`, a Docker image is automatically published to GitHub Packages when the release is tagged.

- https://github.com/igorshubovych/markdownlint-cli/actions/runs/1816284772
- https://github.com/igorshubovych/markdownlint-cli/pkgs/container/markdownlint-cli/versions

PR #212, opened 2021-07-30
Add quiet option

bgrnwd:
This PR adds a quiet option so results are not printed to stdout. 

As an aside, it might be a good idea to add some documentation on contributing/developing locally.

### DIFF
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -212,6 +212,7 @@ program
   .option('-j, --json', 'write issues in json format')
   .option('-o, --output [outputFile]', 'write issues to file (no console)')
   .option('-p, --ignore-path [file]', 'path to file with ignore pattern(s)')
+  .option('-q, --quiet', 'do not write issues to STDOUT')
   .option('-r, --rules  [file|directory|glob|package]', 'custom rule files', concatArray, [])
   .option('-s, --stdin', 'read from STDIN (does not work with files)');
 
@@ -322,7 +323,9 @@ function lintAndPrint(stdin, files) {
   }
 
   const lintResult = markdownlint.sync(lintOptions);
-  printResult(lintResult);
+  if (!options.quiet) {
+    printResult(lintResult);
+  }
 }
 
 if ((files.length > 0) && !options.stdin) {
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -192,7 +192,9 @@ function printResult(lintResult) {
       process.exitCode = 2;
     }
   } else if (lintResultString) {
-    console.error(lintResultString);
+    if(!options.quiet){
+      console.error(lintResultString);
+    }
   }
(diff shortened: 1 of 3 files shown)


markdownlint.js
```diff
@@ -212,6 +212,7 @@ program
   .option('-j, --json', 'write issues in json format')
   .option('-o, --output [outputFile]', 'write issues to file (no console)')
   .option('-p, --ignore-path [file]', 'path to file with ignore pattern(s)')
+  .option('-q, --quiet', 'do not write issues to STDOUT')
```
DavidAnson, line 215: Please update README with this

test/test.js
```diff
@@ -814,3 +814,10 @@ test('without --dot option exclude folders/files with a dot', async t => {
   t.is(result.stdout, '');
   t.is(result.stderr, '');
 });
+
+test('with --quiet option does not print to stdout', async t => {
```
DavidAnson, line 818: Please check for failed exit code

DavidAnson (changes requested):
Minor comments only.

What's the scenario where you don't want to know the errors?

bgrnwd:
> Minor comments only.
> 
> What's the scenario where you don't want to know the errors?

Mainly in conjunction with the `-f` flag. Sometimes I'd just to like fix the errors that are fixable and leave it at that.

test/test.js
DavidAnson: The exit code is wrong here. There are linting errors, but the exit code is 0. It should have been set to 1 in printResult, but this change skipped that call. It may be necessary to duplicate the (results.length > 0) check here.

DavidAnson (changes requested):
Almost there - see my new comment about exit code.

test/test.js
```diff
@@ -814,3 +814,19 @@ test('without --dot option exclude folders/files with a dot', async t => {
   t.is(result.stdout, '');
   t.is(result.stderr, '');
 });
+
+test('with --quiet option does not print to stdout or stderr', async t => {
+  const fixFileQ = 'incorrect.q.mdf';
+  try {
+    fs.copyFileSync('incorrect.md', fixFileQ);
+    await execa('../markdownlint.js',
+      ['--fix', '--quiet','--config', 'test-config.json', fixFileQ],
```
DavidAnson, line 823: Why do we need to use --fix in this test? I think it's simpler if we don't need to copy/edit/delete the file. Though this may raise a good point that --fix should yield exitCode 0 when it's able to fix everything. I don't know if that's currently verified - it's outside the scope of this PR, tho.

DavidAnson (changes requested):
I may see about just editing this in the browser to make the change I'm asking about.

DavidAnson (approved):

PR #234, opened 2021-11-13
Add --enable + --disable CLI flags

### LINKED ISSUE #214: Disable rules from CLI, not config file?
Is it possible to to configure rules directly from the command line without creating a config file? In my case, the ability to disable rules with a flag would be most helpful. Something like

```sh
markdownlint --disable MD013,MD033,MD041 .
```

janosh:
Closes #214.


DavidAnson:
Awesome, thank you!!

Some high-level thoughts:
1. `commander` supports variadic options and using that here should simplify the code a tiny bit and be more consistent: https://github.com/tj/commander.js#variadic-option
2. It would be nice to offer single letter options, but we are running out of letters and 'd' is already taken
3. Please make sure the syntax in README exactly matches what the program outputs
4. No tiny URLs, please, folks can visit the README if they have questions
5. "// leave default values in place if rule is an object" is clever!
6. I think I would prefer to have disable take precedence if a rule is both enabled and disabled at the same time
7. Let's please document number 6 as well - there may not be room in the syntax but it would be good to mention this in the README to call and that command line choices override configuration files
8. I like the tests you've added, could we have one also to validate what happens when the same rule is enabled and disabled
9. There should probably also be a test to validate the implementation of number five above
10. It would also be good to have a test that used the name of the rule to make sure that works as well (ex: "line-length")

janosh:
@DavidAnson Thanks for the quick review. Great comments! I think I addressed all of them.

Two of the tests for `--disable` are failing locally though. Not sure what's happening. I may be misusing the variadic args but not sure how to fix. Perhaps you can spot the issue?

test/test.js
```diff
@@ -851,3 +851,72 @@ test('with --quiet option does not print to stdout or stderr', async t => {
...
+  ['--disable', 'MD002 MD014 MD022 MD041', 'incorrect.md'],
+  {stripFinalNewline: false});
+
+  t.is(result.stdout, '');
+  t.is(result.stderr, '');
+  t.is(result.exitCode, 0)
+
+  try {
+    await execa('../markdownlint.js',
+      ['--disable', 'MD014', 'incorrect.md'],
```
DavidAnson, line 905: I think the test failure is due to incorrect.md being treated as part of the disable list. See https://github.com/tj/commander.js/blob/master/docs/options-taking-varying-arguments.md#alternative-put-options-last for one way to avoid this, tho "--" is also fine and maybe good to show here.

DavidAnson (changes requested):
Looking good!

test/test.js
```diff
@@ -851,3 +851,72 @@ test('with --quiet option does not print to stdout or stderr', async t => {
...
+      {stripFinalNewline: false});
+    // t.fail();
```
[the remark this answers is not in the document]
janosh: Good catch!

test/test.js
janosh: Yeah, I tried adding the `--` in there before but it didn't help. Or rather it wasn't enough. Turns out, you also need to pass each rules as a separate string for them to parse into an array, i.e.
```diff
- ['--disable', 'MD014 MD014 MD022', '--', 'incorrect.md']
+ ['--disable', 'MD014', 'MD014', 'MD022', '--', 'incorrect.md']
```
Do you prefer separating rules with spaces or commas? Spaces avoids the need for a custom parsing function but has slightly worse usability imo.

DavidAnson (approved):
Sweet!

DavidAnson:
By the way, I just now saw notifications for your most recent comments and updates. Sorry about seeming to ignore them, that wasn't the intent!￼

janosh:
@DavidAnson No problem. Do you know when the next release might land?

DavidAnson:
Soon. I did some preparation work last night. :)

Issue #237, opened 2021-11-16
Inessential dependencies


janosh:
@igorshubovych What's your take on getting rid of superseded dependencies? There are a few things here like `lodash.flatten` that could be replaced with standard library alternatives, in this case [`arr.flat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#browser_compatibility) available since `node` v11. Would help with bringing install times and disk usage down.

DavidAnson:
Happy to see this and love the idea as I am a dependency minimalist myself. The minimum supported Node version now is 12 (this should be reflected in package.json?) and my favorite site to check is https://node.green/.

PR #238, opened 2021-11-16
replace some unmaintained/unused node mods with stdlib or own code

janosh:
Adresses #237.

Some easy dependecy removals to get the ball rolling.

Further potential removals:

```json
  "dependencies": {
    "deep-extend": "~0.6.0",
    "get-stdin": "~8.0.0",
    "ignore": "~5.1.9",
    "jsonc-parser": "~3.0.0",
    "minimatch": "~3.0.4",
    "minimist": "~1.2.5",
    "run-con": "~1.2.10"
  },
  "devDependencies": {
    "tap-growl": "^3.0.0",
    "xo": "^0.46.4"
  },
```

Not sure what all of them do. Any you can spot that look outdated / replaceable? If not, will look more into it later.


package.json
```diff
@@ -40,12 +40,9 @@
     "commander": "~8.3.0",
     "deep-extend": "~0.6.0",
     "get-stdin": "~8.0.0",
-    "glob": "~7.2.0",
```
DavidAnson, line 43: I think this is a typo

package.json
```diff
@@ -55,7 +52,6 @@
   "devDependencies": {
     "ava": "^3.15.0",
     "execa": "^5.1.1",
-    "husky": "^7.0.4",
```
DavidAnson, line 58: It's not something I use, but this makes sure precommit runs before commit, even though it is not explicitly referenced.

DavidAnson (commented):
As you see, I made lint part of CI. :)

Looking at the list, maybe deep-extend could also go? I forget what tap-growl does. I'm not sure there's much else you could purge. Great improvements, tho!

package.json
janosh: Typo? I don't understand.

package.json
janosh: Never heard of something like this. Perhaps this should be documented? The readme could have a sentence or two on Husky support.

janosh:
`tap-growl` was apparently used to have Growl notifications on continuous testing:

```json
  "scripts": {
    "invalid": "node ./markdownlint.js --config test/test-config.json -- test/incorrect.md",
    "fix": "cp test/incorrect.md test/incorrect.copy.md && node ./markdownlint.js --fix test/incorrect.copy.md ; cat test/incorrect.copy.md && rm test/incorrect.copy.md",
    "test": "ava",
    "watch": "npm test -- --watch --tap | tap-growl",
    "start": "node ./markdownlint.js",
    "precommit": "xo && npm test"
  },
```

Is [no longer maintained](https://libraries.io/npm/tap-growl) though. Shall we modify the `watch` script to `npm test -- --watch` or get rid of it enirely?

package.json
janosh: Reinstalled husky. Dropped minimist as it didn't seem to serve any function. All test pass with or without it.

janosh:
Ready for review.

package.json
DavidAnson: glob is still used.

package.json
DavidAnson: Husky is only for developers. I'd be fine if you remove this. I added xo to CI because so many people weren't running it. :)

package.json
DavidAnson: It's a direct dependency. Try running the code after "npm install --production".

DavidAnson:
I'll look over the rest of the changes soon, but I think I will wait to approve this PR until after the next release which I hope to do in a couple of days. I don't want to risk some surprise here breaking things.

package.json
janosh: Oh right, for some reason I was convinced newer versions of NodeJS come with a 1st party glob module like Python.

janosh:
Good plan!

package.json
```diff
@@ -38,25 +38,19 @@
   "homepage": "https://github.com/igorshubovych/markdownlint-cli#readme",
   "dependencies": {
     "commander": "~8.3.0",
-    "deep-extend": "~0.6.0",
     "get-stdin": "~8.0.0",
-    "glob": "~7.2.0",
+    "glob": "^7.2.0",
```
DavidAnson, line 42: Did you mean to change tilde to caret?

markdownlint.js
```diff
@@ -38,7 +35,7 @@ const processCwd = process.cwd();
 
 function readConfiguration(userConfigFile) {
   const jsConfigFile = /\.js$/i.test(userConfigFile);
-  const rcArgv = minimist(process.argv.slice(2));
```
DavidAnson, line 41: I was worried about this change, but it looks like the result wasn't used. There is no third parameter to run-con, could you remove it a few lines below, please?

https://www.npmjs.com/package/run-con

DavidAnson (changes requested):
Nice improvements! Two comments, both easy.

markdownlint.js
janosh: Good catch.

package.json
janosh: Oops, not intentional.

markdownlint.js
```diff
@@ -38,29 +35,27 @@ const processCwd = process.cwd();
...
+  const rcArgv = process.argv.slice(2);
```
DavidAnson, line 38: Not easy to tell from my phone, but I think rcArgv is no longer used?

markdownlint.js
```diff
@@ -38,29 +35,27 @@ const processCwd = process.cwd();
...
   if (jsConfigFile) {
     // Prevent rc package from parsing .js config file as INI
     delete rcArgv.config;
   }
 
   // Load from well-known config files
-  let config = rc('markdownlint', {}, rcArgv);
+  let config = rc('markdownlint', {});
```
DavidAnson, line 45: Thanks!

DavidAnson (changes requested):
One last question!

markdownlint.js
```diff
@@ -10,13 +10,10 @@ const process = require('process');
 const program = require('commander');
 
 const options = program.opts();
-const differenceWith = require('lodash.differencewith');
-const flatten = require('lodash.flatten');
+const glob = require('glob');
 const markdownlint = require('markdownlint');
 const rc = require('run-con');
-const glob = require('glob');
```
nschonni, line 17: minor, but you might want to move the require back to the same position to reduce the diff.
Doesn't look like there was a particular order to start with though

DavidAnson: I agree in general, but this one seems harmless.

DavidAnson (approved):
Awesome!

PR #245, opened 2021-12-10
Standardize and improve exit code handling

sxlijin:
This PR does two things:

  1.  standardize on:

      * `exit(1)` when markdownlint runs successfully but finds issues in the input and

      * `exit(2)` when markdownlint fails to run successfully, perhaps because a configuration file is malformed or perhaps because some kind of filesystem error has occurred

  2.  handle errors more robustly: a runtime error inside `lintAndPrint()` today results in `exit(1)`, e.g. for a file that we don't have read permissions on (note: this is not comprehensive, we do still do some code execution before `lintAndPrint()`, but this is still a solid improvement)

The background for this PR is that I'm working on [a universal linter](https;//trunk.io/products/check) which makes it easy for users to run linters for different subsets of their repos, and `markdownlint-cli` uses `exit(1)` as a catch-all for not just lint findings but also unexpected failures.

Specifically we've seen `markdownlint` return with `exit(1)` when this happens:

```
internal/fs/utils.js:314
    throw err;
    ^

Error: ENOENT: no such file or directory, open '[sanitized]'
    at Object.openSync (fs.js:498:3)
    at Object.readFileSync (fs.js:394:35)
    at lintFile ([sanitized]/.cache/trunk/linters/markdownlint/0.29.0-ee485b2571c3867cdf56d258aa9aed05/node_modules/markdownlint/lib/markdownlint.js:714:33)
    at lintInput ([sanitized]/.cache/trunk/linters/markdownlint/0.29.0-ee485b2571c3867cdf56d258aa9aed05/node_modules/markdownlint/lib/markdownlint.js:790:7)
    at Function.markdownlintSync [as sync] ([sanitized]/.cache/trunk/linters/markdownlint/0.29.0-ee485b2571c3867cdf56d258aa9aed05/node_modules/markdownlint/lib/markdownlint.js:888:3)
    at lintAndPrint ([sanitized]/.cache/trunk/linters/markdownlint/0.29.0-ee485b2571c3867cdf56d258aa9aed05/node_modules/markdownlint-cli/markdownlint.js:324:35)
    at Object.<anonymous> ([sanitized]/.cache/trunk/linters/markdownlint/0.29.0-ee485b2571c3867cdf56d258aa9aed05/node_modules/markdownlint-cli/markdownlint.js:329:3)
    at Module._compile (internal/modules/cjs/loader.js:1072:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1101:10)
    at Module.load (internal/modules/cjs/loader.js:937:32) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '[sanitized]'
}
```

and this can be easily reproduced via:

```
touch foo.md
chmod a-r foo.md
markdownlint-cli foo.md
```

### DIFF
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -339,11 +339,9 @@ function main() {
   }
 }
 
-if (require.main === module) {
-  try {
-    main();
-  } catch (error) {
-    console.error(error);
-    process.exit(exitCodes.unexpectedError);
-  }
+try {
+  main();
+} catch (error) {
+  console.error(error);
+  process.exit(exitCodes.unexpectedError);
 }
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -24,7 +24,12 @@ function jsYamlSafeLoad(text) {
   return require('js-yaml').load(text);
 }
 
-const exitCodes = { lintFindings: 1, unexpectedError: 2 };
+const exitCodes = {
+  lintFindings: 1,
+  failedToWriteOutput: 2,
+  failedToLoadCustomRules: 3,
+  unexpectedError: 4
+};
 
 const projectConfigFiles = [
   '.markdownlint.json',
@@ -241,7 +246,7 @@ function loadCustomRules(rules) {
       return fileList;
     } catch (error) {
       console.error('Cannot load custom rule ' + rule + ': ' + error.message);
-      return process.exit(3);
+      return process.exit(exitCodes.failedToLoadCustomRules);
     }
   });
 }
@@ -339,5 +344,6 @@ if (require.main === module) {
     main();
   } catch (error) {
     process.exitCode = exitCodes.unexpectedError;
+    throw error;
   }
 }
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -328,7 +328,7 @@ function lintAndPrint(stdin, files) {
   printResult(lintResult);
 }
 
-function main() {
+try {
   if ((files.length > 0) && !options.stdin) {
     lintAndPrint(null, diff);
   } else if ((files.length === 0) && options.stdin && !options.fix) {
@@ -337,10 +337,6 @@ function main() {
   } else {
     program.help();
   }
-}
-
-try {
-  main();
 } catch (error) {
   console.error(error);
   process.exit(exitCodes.unexpectedError);
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -24,6 +24,8 @@ function jsYamlSafeLoad(text) {
   return require('js-yaml').load(text);
 }
 
+const exitCodes = { lintFindings: 1, unexpectedError: 2 };
+
 const projectConfigFiles = [
   '.markdownlint.json',
   '.markdownlint.yaml',
@@ -61,7 +63,7 @@ function readConfiguration(userConfigFile) {
       config = require('deep-extend')(config, userConfig);
     } catch (error) {
       console.error(`Cannot read or parse config file '${userConfigFile}': ${error.message}`);
-      process.exitCode = 1;
+      process.exitCode = exitCodes.unexpectedError;
     }
   }
 
@@ -160,7 +162,7 @@ function printResult(lintResult) {
     // and let the program terminate normally.
     // @see {@link https://nodejs.org/dist/latest-v8.x/docs/api/process.html#process_process_exit_code}
     // @see {@link https://github.com/igorshubovych/markdownlint-cli/pull/29#issuecomment-343535291}
-    process.exitCode = 1;
+    process.exitCode = exitCodes.lintFindings;
   }
 
   if (options.output) {
@@ -171,7 +173,7 @@ function printResult(lintResult) {
       fs.writeFileSync(options.output, lintResultString);
     } catch (error) {
       console.warn('Cannot write to output file ' + options.output + ': ' + error.message);
-      process.exitCode = 2;
+      process.exitCode = exitCodes.unexpectedError;
     }
   } else if (lintResultString && !options.quiet) {
     console.error(lintResultString);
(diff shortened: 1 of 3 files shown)


DavidAnson (changes requested):
Thanks for the contribution!

It feels like this changes the meaning of exit code 2 from what is documented: https://github.com/igorshubovych/markdownlint-cli#exit-codes

Perhaps you want to create a new error code to distinguish these cases from code 1?

FYI, there seem to be issues breaking CI.

Unrelated, you may want to look at the following as it seems similar to your project: https://github.com/github/super-linter

sxlijin:
> It feels like this changes the meaning of exit code 2 from what is documented: https://github.com/igorshubovych/markdownlint-cli#exit-codes
>
> Perhaps you want to create a new error code to distinguish these cases from code 1?

Ah, I somehow looked for this everywhere except in the README!

A new error code certainly makes sense.

I am slightly concerned about `exit(1)` being overloaded for lint findings and for a bad config though. Would you be open to splitting those error codes?

> FYI, there seem to be issues breaking CI.

Ack, opened this and then went to look at something else and didn't see the CI failures. Will keep an eye on these.

> Unrelated, you may want to look at the following as it seems similar to your project: https://github.com/github/super-linter

Thanks for the heads-up. I should clarify that this isn't just a personal project but we're actually a company and this is one of the products that we're building out, so this is something we're very aware of :)

markdownlint.js
```diff
@@ -24,6 +24,13 @@ function jsYamlSafeLoad(text) {
   return require('js-yaml').load(text);
 }
 
+const exitCodes = {
```
DavidAnson, line 27: Nice!

markdownlint.js
```diff
@@ -24,6 +24,13 @@ function jsYamlSafeLoad(text) {
...
+  lintFindings: 1,
```
DavidAnson, line 28: "lintFailures", maybe?

markdownlint.js
```diff
@@ -321,11 +328,22 @@ function lintAndPrint(stdin, files) {
...
+    lintAndPrint(null, diff);
+  } else if ((files.length === 0) && options.stdin && !options.fix) {
+    const getStdin = require('get-stdin');
+    getStdin().then(lintAndPrint);
+  } else {
+    program.help();
+  }
+}
+
+if (require.main === module) {
```
DavidAnson, line 342: Why add this?

test/test.js
```diff
@@ -126,6 +126,22 @@ test('linting of incorrect Markdown file fails with absolute path', async t => {
   }
 });
 
+test('linting of unreadable Markdown file fails', async t => {
+  const unreadablePath = '/tmp/unreadable.md';
```
DavidAnson, line 130: Could you do relative to __dirname? I don't want to reach out like this as some systems may treat that as malware, etc..

test/test.js
```diff
@@ -126,6 +126,22 @@ test('linting of incorrect Markdown file fails with absolute path', async t => {
...
+  try {
+    await execa('../markdownlint.js',
+      ['--config', 'test-config.json', unreadablePath],
+      {stripFinalNewline: false});
+    t.fail();
+  } catch (error) {
+    t.is(error.exitCode, 4);
+  }
+
+  fs.rmSync(unreadablePath, {force: true});
```
DavidAnson, line 142: Put this in a finally?

DavidAnson (changes requested):
Left a few comments/questions on the changes.

test/test.js
DavidAnson: Also, it breaks on Windows. :)

markdownlint.js
sxlijin: Ah, this is my Python background leaking out. I guess it doesn't really matter if you never expect anyone to import this as a module. Will lift this out.

test/test.js
sxlijin: Eurghhh. This was causing race issues with the other tests, that's why I went with `/tmp` originally. Need to figure out somewhere it doesn't break the glob tests.

markdownlint.js
sxlijin: "lintFailures" suggests to me that linting failed to happen, aka an unexpected failure, hence "lintFindings"

This one's a tricky terminology issue and is also frustratingly non standard across a lot of linters :(

DavidAnson: lintIssues?

test/test.js
DavidAnson: You can give it a non-standard extension? .mdtest or something?

markdownlint.js
```diff
@@ -321,11 +328,20 @@ function lintAndPrint(stdin, files) {
...
 }
 
-if ((files.length > 0) && !options.stdin) {
-  lintAndPrint(null, diff);
-} else if ((files.length === 0) && options.stdin && !options.fix) {
-  const getStdin = require('get-stdin');
-  getStdin().then(lintAndPrint);
-} else {
-  program.help();
+function main() {
```
DavidAnson, line 331: What do you think about just putting this inside the try?

test/test.js
```diff
@@ -126,6 +126,24 @@ test('linting of incorrect Markdown file fails with absolute path', async t => {
   }
 });
 
+test('linting of unreadable Markdown file fails', async t => {
+  const unreadablePath = '../unreadable.test.md';
+  try {
+    fs.writeFileSync(unreadablePath, '', {mode: 0o222});
+
+    try {
```
DavidAnson, line 134: Not sure we need the nested try? I think the outer can have a catch and a finally?

DavidAnson (changes requested):
Couple more questions, replies to a couple of comments.

DavidAnson:
Looks like 2 CI issues:
1. message: 'fs.rmSync is not a function' on Node 12
2. Unreadable file is readable on Windows?

DavidAnson:
`mode` seems not to be well supported on Windows, see last paragraph: https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fschmodpath-mode-callback

markdownlint.js
sxlijin: done!

sxlijin:
maybe a deliberately broken symlink will work on windows?

DavidAnson (approved):
Thanks!

sxlijin:
and thanks to you for being so quick and responsive!

PR #247, opened 2021-12-14
Set xo to prettier compat mode + auto fixes

### LINKED ISSUE #246: Prettier config
Right now, when using an IDE with format on save, you get a lot of changes when saving `.js` files. How about we add `.prettierrc.yml` to ensure consistent formatting?

janosh:
Closes #246.

Add `.prettierrc.yml` and apply it to all `.js` files with `npx prettier --write **/*.js`.

I tried to find the config that resulted in the minimal amount of changes. Happy to tweak it some more though. For instance, I set `printWidth: 88` following the Python formatter `black` but JS sometimes uses different values.

### DIFF
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -2,3 +2,4 @@ semi: true
 singleQuote: true
 bracketSpacing: false
 endOfLine: auto
+arrowParens: avoid
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -4,3 +4,4 @@ bracketSpacing: false
 endOfLine: auto
 arrowParens: avoid
 printWidth: 88
+trailingComma: none
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -3,3 +3,4 @@ singleQuote: true
 bracketSpacing: false
 endOfLine: auto
 arrowParens: avoid
+printWidth: 88
--- a/.prettierrc.yml
+++ b/.prettierrc.yml
@@ -1,0 +1,4 @@
+semi: true
+singleQuote: true
+bracketSpacing: false
+endOfLine: auto
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -86,7 +86,7 @@ function prepareFileList(files, fileExtensions, previousResults) {
     extensionGlobPart += '{' + fileExtensions.join(',') + '}';
   }
 
-  files = files.map((file) => {
+  files = files.map(file => {
     try {
       if (fs.lstatSync(file).isDirectory()) {
         // Directory (file falls through to below)
@@ -96,8 +96,8 @@ function prepareFileList(files, fileExtensions, previousResults) {
             globOptions,
           );
           return previousResults
-            .filter((fileInfo) => matcher.match(fileInfo.absolute))
-            .map((fileInfo) => fileInfo.original);
+            .filter(fileInfo => matcher.match(fileInfo.absolute))
+            .map(fileInfo => fileInfo.original);
         }
 
         return glob.sync(path.join(file, '**', extensionGlobPart), globOptions);
@@ -110,8 +110,8 @@ function prepareFileList(files, fileExtensions, previousResults) {
           globOptions,
         );
         return previousResults
-          .filter((fileInfo) => matcher.match(fileInfo.absolute))
-          .map((fileInfo) => fileInfo.original);
+          .filter(fileInfo => matcher.match(fileInfo.absolute))
+          .map(fileInfo => fileInfo.original);
       }
 
       return glob.sync(file, globOptions);
@@ -120,7 +120,7 @@ function prepareFileList(files, fileExtensions, previousResults) {
     // File
     return file;
   });
-  return files.flat().map((file) => ({
+  return files.flat().map(file => ({
     original: file,
     relative: path.relative(processCwd, file),
     absolute: path.resolve(file),
@@ -128,8 +128,8 @@ function prepareFileList(files, fileExtensions, previousResults) {
 }
 
 function printResult(lintResult) {
-  const results = Object.keys(lintResult).flatMap((file) =>
-    lintResult[file].map((result) => {
+  const results = Object.keys(lintResult).flatMap(file =>
+    lintResult[file].map(result => {
       if (options.json) {
         return {
           fileName: file,
@@ -172,7 +172,7 @@ function printResult(lintResult) {
       );
 
       lintResultString = results
-        .map((result) => {
+        .map(result => {
           const {file, lineNumber, column, names, description} = result;
           const columnText = column ? `:${column}` : '';
           return `${file}:${lineNumber}${columnText} ${names} ${description}`;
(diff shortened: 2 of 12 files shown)


DavidAnson:
Yes, I prefer to enforce style rules with a linter. As you note, xo is already used by this project and that is a nice way to do this. What's nice about this approach is that it is automatically enforced by CI, so even someone who doesn't know to run the right tool will be informed as part of their PR. I would prefer to enable one rule at a time so the differences are easier to understand and discuss. If people want to use an editor or extension that honors these settings, more power to them - I don't, tho, and do not want to require that.

package.json
```diff
@@ -15,7 +15,7 @@
     "test": "ava",
     "watch": "npm test --watch",
     "start": "node ./markdownlint.js",
-    "precommit": "xo && npm test"
+    "precommit": "xo --prettier && npm test"
```
nschonni, line 18: Shouldn't the boolean be set down below on line 57 config object instead of the cli?

janosh: Oops, didn't even see that config.

markdownlint.js
```diff
@@ -54,13 +57,15 @@ function readConfiguration(userConfigFile) {
   if (userConfigFile) {
     try {
       const userConfig = jsConfigFile
-        // Evaluate .js configuration file as code
-        ? require(path.resolve(processCwd, userConfigFile))
-        // Load JSON/YAML configuration as data
-        : markdownlint.readConfigSync(userConfigFile, configFileParsers);
+        ? // Evaluate .js configuration file as code
```
DavidAnson, line 60: This is unusual

DavidAnson (changes requested):
Thanks for updating to use xo! I'd still love to see this added in stages (infra, rule 1, rule 2, etc.) so the diff is easier to review. I also left a comment where I think auto-fix did something weird.

markdownlint.js
janosh: Agreed. I manually reformatted this in the latest commit series.

janosh:
> Thanks for updating to use xo! I'd still love to see this added in stages (infra, rule 1, rule 2, etc.) so the diff is easier to review.

Happy to do it. Just bear in mind that this leads to intermittent changes (e.g. from trailing commas that are inserted when adding the `--prettier` flag to `xo` and then removed again when setting `trailingComma: none` in `.prettierrc.yml`.

DavidAnson:
I'm happy to have the intermittent changes - right now, GitHub won't even render `test.js` because there are too many edits. :) Assuming you're okay with that, I'll look for an update here that adds the infrastructure and enables a single rule?

janosh:
> Assuming you're okay with that, I'll look for an update here that adds the infrastructure and enables a single rule?

The latest series of commits is not what you were asking? I must have misunderstood. Could you explain what you mean by "adds the infrastructure and enables a single rule?"?

.prettierrc.yml
```diff
@@ -0,0 +1,7 @@
+semi: true
```
nschonni, line 1: It seems like the recomendation from https://github.com/xojs/xo#prettier is to just set 
```
	trailingComma: 'es5',
	singleQuote: false,
	bracketSpacing: true,
```
Not sure about some of the others.

janosh: Happy to use that instead. Like I said, was trying to find a config that minimizes changes.

DavidAnson:
The way I read `prettier.yml` now, it is configuring and enforcing 7 style rules. That leads to so many changes to `test.js` that GitHub won't render it. I'm proposing enabling just a single rule at first, we review that PR and get it merged, then send a new PR enabling another rule (maybe two), PR, merge, etc.. Its a bit more process, but we have more confidence in each step along the way because the number of changes is smaller.

janosh:
> The way I read `prettier.yml` now, it is configuring and enforcing 7 style rules

But the additional rules result in less changes. If I leave for example

```yml
arrowParens: avoid
trailingComma: none
```

there would be an even larger diff. Not sure what we'd gain by splitting into multiple PRs. This way at least you have effects of adding Prettier in one place.

DavidAnson:
Okay, thanks, I'll look into this when I have some time.

janosh:
@DavidAnson What's your view on the rule set? Minimal formatting changes (requiring more rules) or minimal rule set as @nschonni suggested?

DavidAnson:
I don't mind a strict end state (I usually enable as many linting rules as I can), but I'd like to get there incrementally as much as possible. If that means the rule set definition changes each step along the way, no concerns from me. The code itself (and tests) is what I prioritize not introducing errors into.

janosh:
> The code itself (and tests) is what I prioritize not introducing errors into.

Understandable though both Prettier and this repo are so well tested that broken functionality from auto-formatting seems unlikely.

DavidAnson:
From what I can tell (not a Prettier user), Prettier is not configurable enough to allow an easy rule-by-rule migration. Attempting to configure it for 0 issues with the current code as-is, I don't see how to get fewer than 222 errors. The following configuration is the best I found so far:

```diff
diff --git a/package.json b/package.json
index eb76b18..fd68447 100644
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "xo": "^0.47.0"
   },
   "xo": {
+    "prettier": true,
     "space": true,
     "rules": {
       "comma-dangle": 0,
@@ -67,6 +68,11 @@
       "unicorn/prefer-ternary": 0
     }
   },
+  "prettier": {
+    "arrowParens": "avoid",
+    "endOfLine": "auto",
+    "trailingComma": "none"
+  },
   "ava": {
     "files": [
       "test/**/*.js",
```

Prettier doesn't seem to have a way to disable its indent rule and it has very strong feelings about line-breaks. It insists on `Replace (files.length·>·0) with files.length·>·0` for a line like `if ((files.length > 0) && !options.stdin) {` which I fundamentally disagree with.

I am still experimenting, but the tool is not winning me over.

DavidAnson:
Okay, by setting `printWidth` to 1000, I avoided many of the wrapping issues. This commit represents the minimal diff I got to: https://github.com/igorshubovych/markdownlint-cli/commit/d62993f6757bea0e27fffd60ed9f941cdb8314ba. I'll look over it tomorrow to form an opinion. Annoyingly, `printWidth` will seemingly reformat lines that have been carefully formatted for readability and jam them onto a single line (see `projectConfigFiles`) - which seems to defeat efforts at clarity in favor of absolute consistency.

nschonni:
Not ideal, but you can ignore specific blocks https://prettier.io/docs/en/ignore.html

janosh:
@DavidAnson In Prettier's defense, they do call themselves "An opinionated code formatter". There are certainly cases where it produces non-ideal formatting. For me, the benefit of not having to do the formatting manually outweighs that drawback. If it's not at all to your liking, feel free to close this PR.

janosh:
@DavidAnson `package.json` needs two more lines for files to remain unchanged by prettier formatting (which was most the goal behind this PR):

```json
  "prettier": {
    "arrowParens": "avoid",
    "endOfLine": "auto",
    "printWidth": 1000,
    "trailingComma": "none",
+   "singleQuote": true,
+   "bracketSpacing": false
  },
```

Also, just my 2 cents but I think those super long lines from `pinrtWidth: 1000` aren't better than what `prettier` produces by default.

PR #254, opened 2022-01-10
chore: disable Ava Workers

nschonni:
See if it resolves the failure for process.chdir with Ava v4


nschonni:
@DavidAnson not sure why GitHub decided to cancel those 2 macOS builds, but it does look like it resolves it. My local failure was on Windows through git bash, so maybe something else going on

PR #268, opened 2022-03-02
.pre-commit-hooks.yaml: add markdownlint-docker

macpijan:
It can be useful to run the tool via container for pre-hook if one does
not feel like installing node on the host system.

Signed-off-by: Maciej Pijanowski <maciej.pijanowski@3mdeb.com>

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -18,3 +18,9 @@
     entry: ghcr.io/igorshubovych/markdownlint-cli
     language: docker_image
     types: [markdown]
+-   id: markdownlint-fix-docker
+    name: markdownlint-fix-docker
+    description: "Fixes the style of Markdown/Commonmark files, uses projects official docker image.."
+    entry: ghcr.io/igorshubovych/markdownlint-cli --fix
+    language: docker_image
+    types: [markdown]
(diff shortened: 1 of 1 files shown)


.pre-commit-hooks.yaml
```diff
@@ -12,3 +12,15 @@
     language: node
     types: [markdown]
     minimum_pre_commit_version: 0.15.0
+-   id: markdownlint-docker
+    name: markdownlint-docker
+    description: "Checks the style of Markdown/Commonmark files, uses projects official docker image."
+    entry: ghcr.io/igorshubovych/markdownlint-cli
+    language: docker_image
+    types: [markdown]
```
DavidAnson, line 20: Please add "minimum_pre_commit_version".

.pre-commit-hooks.yaml
```diff
@@ -12,3 +12,15 @@
     language: node
     types: [markdown]
     minimum_pre_commit_version: 0.15.0
+-   id: markdownlint-docker
+    name: markdownlint-docker
+    description: "Checks the style of Markdown/Commonmark files, uses projects official docker image."
```
DavidAnson, line 17: I think the part after the comma is not needed - using the same description here seems fine.

DavidAnson (changes requested):
Didn't know this was possible!

.pre-commit-hooks.yaml
macpijan: Changed

.pre-commit-hooks.yaml
macpijan: Added

DavidAnson (approved):
Thanks!

PR #270, opened 2022-03-05
feat: ✨  Add support for `markdownlint.cjs` config in ESM workspace

### LINKED ISSUE #269: Support JS configuration in ESM workspaces (*.cjs)
## The problem

In the workspace with option `"type": "module"` set inside the `package.json` file at the project root. When you have a JavaScript configuration file, you cannot use flag `--config "./.markdownlint.js"`, because it will throw the following error:

```
> markdownlint . --config "./.markdownlint.js" --dot --ignore-path "./.gitignore" --ignore "./{.changeset,helpers,packages}/**/*"

Cannot read or parse config file './.markdownlint.js': require() of ES Module /***/.markdownlint.js from /***/node_modules/.pnpm/markdownlint-cli@0.31.1/node_modules/markdownlint-cli/markdownlint.js not supported.
.markdownlint.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename .markdownlint.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /***/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

Renaming to the `.markdownlint.cjs` and modyfing the script flag to `--config "./.markdownlint.cjs"` will return this error:

```
Cannot read or parse config file './.markdownlint.cjs': Unable to parse './.markdownlint.cjs'; Parser 0: Unexpected token m in JSON at position 0; Parser 1: end of the stream or a document separator is expected (2:15)

 1 | module.exports = {
 2 | →"line-length": false,
-------------------^
 3 | }
 ```

### PR DESCRIPTION
**Pull request by xeho91**
Resolves #269

### CODE REVIEW

File: markdownlint.js [hunk 1]
```diff
@@ -41,7 +41,7 @@ const fsOptions = {encoding: 'utf8'};
 const processCwd = process.cwd();
 
 function readConfiguration(userConfigFile) {
-  const jsConfigFile = /\.js$/i.test(userConfigFile);
+  const jsConfigFile = /\.(c)?js$/i.test(userConfigFile);
```
DavidAnson, line 44: I think the parentheses are unnecessary?

DavidAnson (changes requested):
Cool! Could you please also add a test case that uses a .cjs file?

DavidAnson:
... and update the README/help if applicable.

xeho91:
> Cool! Could you please also add a test case that uses a .cjs file?

Done.

I am not sure if this is the correct testing strategy. I am not experienced in writing tests. I would appreciate any guidance if this needs to change.

xeho91:
> ... and update the README/help if applicable.

Done. Is it straightforward enough?

markdownlint.js
xeho91: You are right. I have a habit of always wrapping with parentheses for any optional regexes. :slightly_smiling_face:

README.md
```diff
@@ -105,14 +105,17 @@ The CLI argument `--config` is not required.
 If it is not provided, `markdownlint-cli` looks for the file `.markdownlint.json`/`.markdownlint.yaml`/`.markdownlint.yml` in current folder, or for the file `.markdownlintrc` in the current or all parent folders.
 The algorithm is described in detail on the [`rc` package page][rc-standards].
 If the `--config` argument is provided, the file must be valid JSON, JSONC, JS, or YAML.
-JS configuration files contain JavaScript code, must have the `.js` extension, and must export (via `module.exports = ...`) a configuration object of the form shown above.
+JS configuration files contain JavaScript code, must have the `.(c)js` extension, and must export (via `module.exports = ...`) a configuration object of the form shown above.
```
DavidAnson, line 108: Nice. For clarity, could you change to ".js or .cjs" so people don't need to understand parentheses?

README.md
```diff
@@ -105,14 +105,17 @@ The CLI argument `--config` is not required.
...
+So, if your workspace _(project)_ is [ESM-only] _(`"type": "module"`set in the root 'package.json' file)_, then the configuration file **should end with `.cjs` file extension**.
```
DavidAnson, line 109: Great. Maybe remove "So,"?

README.md
```diff
@@ -105,14 +105,17 @@ The CLI argument `--config` is not required.
...
 A JS configuration file may internally `require` one or more npm packages as a way of reusing configuration across projects.
 
 `--enable` and `--disable` override configuration files; if a configuration file disables `MD002` and you pass `--enable MD002`, it will be enabled.
 If a rule is passed to both `--enable` and `--disable`, it will be disabled.
 
 > JS configuration files must be provided via the `--config` argument; they are not automatically loaded because running untrusted code is a security concern.
 
+[ESM-only]:
```
DavidAnson, line 117: Missing link?

test/test.js
```diff
@@ -786,3 +786,17 @@ test('--disable flag overrides --enable flag', async t => {
...
 });
+
+test('configuration can be .js in the CommonJS workspace', async t => {
+  const result = await execa('../../../markdownlint.js', ['--config', '.markdownlint.js', 'test.md'], {cwd: "./workspace/commonjs", stripFinalNewline: false});
+  t.is(result.stdout, '');
+  t.is(result.stderr, '');
+  t.is(result.exitCode, 0);
+});
+
+test('configuration can be .cjs in the ESM (module) workspace', async t => {
```
DavidAnson, line 797: CJS should work in both scenarios, right? Could you please add a third test to confirm that?

DavidAnson (changes requested):
Love the new tests!

DavidAnson:
BTW: "npm run precommit" to see the CI errors locally.

test/test.js
xeho91: Done. It does indeed works in both scenarios. :slightly_smiling_face:

README.md
xeho91: Fixed. I added a link to the popular gist, which probably has all the answers regarding ESM vs CJS workspace and usage in Node.js projects.

README.md
xeho91: Removed.

README.md
xeho91: Changed.

DavidAnson (approved):

Issue #292, opened 2022-06-04
How to run markdownlint-cli recursively?


iamtodor:
Hello,

I have the following structure:
```
tree -L 2
.
├── README.md
├── task1
│   ├── README.md
├── task2
│   ├── README.md
├── task3
│   ├── README.md
└── task4
    ├── README.md
```
Is it possible to run this tool recursively? 
I'm able to run the check against the top README
```
>>> docker run -v $PWD:/workdir ghcr.io/igorshubovych/markdownlint-cli:latest "*.md"
```
But I cannot run it for deeper files:
```
>>> docker run -v $PWD:/workdir ghcr.io/igorshubovych/markdownlint-cli:latest "task1/*.md"
Usage: markdownlint [options] <files|directories|globs>

MarkdownLint Command Line Interface
```
I'd like to the check against all the markdown files in the dir

nschonni:
You're missing the first part of the folder wildcard if you want it have a recursive glob `"**/*.md"`

DavidAnson:
Nick is right, though I would expect your second example to work if there are .md files in the task1 directory.

iamtodor:
@nschonni @DavidAnson great, thank you!
In this situation, when you have the following project (files) structure when all the files have the same name `README.md` would it be possible to provide all the path to the file? 
Currently, it shows only file name:
```
docker run -v $PWD:/workdir ghcr.io/igorshubovych/markdownlint-cli:latest "**/*.md"    
README.md:1 MD022/blanks-around-headings/blanks-around-headers Headings should be surrounded by blank lines [Expected: 1; Actual: 0; Below] [Context: "# Prerequisites"]
README.md:2:81 MD013/line-length Line length [Expected: 80; Actual: 188]
README.md:6:81 MD013/line-length Line length [Expected: 80; Actual: 100]
README.md:18:81 MD013/line-length Line length [Expected: 80; Actual: 123]
README.md:22:81 MD013/line-length Line length [Expected: 80; Actual: 178]
README.md:23:81 MD013/line-length Line length [Expected: 80; Actual: 216]
README.md:25:81 MD013/line-length Line length [Expected: 80; Actual: 243]
```
However, I know it's about 3 different files. 

So what I'd like to see is:
```
docker run -v $PWD:/workdir ghcr.io/igorshubovych/markdownlint-cli:latest "**/*.md"    
[...]

PR #293, opened 2022-06-05
Update README.md

iamtodor:
add a clarification for https://github.com/igorshubovych/markdownlint-cli/issues/292

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Quoting the glob like `--ignore '*.md'` passes it through unexpanded and ignores
 
 #### Globbing examples
 
-To lint all Markdown files recursively in a Node.js project (excluding dependencies), `*/**.md` wildcard comes in handy. The following commands might be used:
+To lint all Markdown files recursively in a Node.js project (excluding dependencies), `**/*.md` wildcard comes in handy. The following commands might be used:
 
 Windows CMD: `markdownlint **/*.md --ignore node_modules`
 
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Quoting the glob like `--ignore '*.md'` passes it through unexpanded and ignores
 
 #### Globbing examples
 
-To lint all Markdown files in a Node.js project (excluding dependencies), the following commands might be used:
+To lint all Markdown files recursively in a Node.js project (excluding dependencies), `*/**.md` wildcard comes in handy. The following commands might be used:
 
 Windows CMD: `markdownlint **/*.md --ignore node_modules`
 


README.md
```diff
@@ -57,7 +57,7 @@ Quoting the glob like `--ignore '*.md'` passes it through unexpanded and ignores
 
 #### Globbing examples
 
-To lint all Markdown files in a Node.js project (excluding dependencies), the following commands might be used:
+To lint all Markdown files recursively in a Node.js project (excluding dependencies), `*/**.md` wildcard comes in handy. The following commands might be used:
```
DavidAnson, line 60: I don't think `*/**.md` is what's intended? And `**/*.md` is already shown on the next two lines.

DavidAnson (changes requested):
Sorry, don't understand this?

README.md
iamtodor: @DavidAnson my intention was to add mention the explanation of this wildcard that is used in the code

DavidAnson: I think lines 62 and 64 already show the preferred syntax?

iamtodor: @DavidAnson yeah, I agree it shows the syntax. 
So I didn't change the syntax, but rather add a precise clarification as a word explanation about the syntax, right?

iamtodor: @DavidAnson what do you think of these changes? Is there anything that concerns you that we can discuss?

DavidAnson: As I wrote above, I think "*/**" is wrong and I think the fact this example is shown on the next two lines is probably already good. I think the phrase "lint ... recursively" may be confusing as there is no recursion from the user's perspective.

iamtodor: @DavidAnson okay, thank you for your sharing!
Here are my thoughts, that come from the issue https://github.com/igorshubovych/markdownlint-cli/issues/292
When I tried to search for this logic in the readme I used the `cmd+f` search on the page with the following `wildcard` and `recursively`. 

I don't quite understand why `I think "*/**" is wrong` because here https://github.com/igorshubovych/markdownlint-cli/issues/292#issuecomment-1146696384 this is precisely what was mentioned

iamtodor: @DavidAnson If you still think this PR has no sense, and it's not discussable please let me know - I will close it :)

DavidAnson: Please read the comment you link to again, it does not say what you think it says.

iamtodor: @DavidAnson `You're missing the first part of the folder wildcard if you want it have a recursive glob "**/*.md"` ? taken from https://github.com/igorshubovych/markdownlint-cli/issues/292#issuecomment-1146696384

DavidAnson: The existing examples and Nick's comment use the syntax `**/*.md` which is the conventional way to match all files with a particular extension. Your proposed change uses the syntax `*/**.md` which means something different.

iamtodor: @DavidAnson ah, sorry, it was a typo. I didn't notice it. I thought to add `*/**.md`, but mistakenly I added `**/*.md`. Thank you for your look, I updated the PR accordingly

Please, let me know if it works for you :)

Realprach (approved):

DavidAnson:
Closing old PR.

PR #315, opened 2022-08-16
Update the help for `--enable` & `--disable`

MFTabriz:
to reflect 33f319cf6ae051d0634536475dcd9b7fcd974b9d

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,8 @@ markdownlint --help
     -q, --quiet                                 do not write issues to STDOUT
     -r, --rules  [file|directory|glob|package]  include custom rule files (default: [])
     -s, --stdin                                 read from STDIN (does not work with files)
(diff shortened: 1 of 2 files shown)


DavidAnson:
I considered this when I made the related change, but did not do it because I feel like it clutters up the list of options. I felt like that list should be more about what the options are instead of general syntax. For example, the two dashes are not needed when the list is the final parameter. I don't feel very strongly about this, but as it's only come up once so far I figured that calling it out in the README would be sufficient.

DavidAnson:
What about just showing the two dashes in the example like you have but not adding the text talking about them?

MFTabriz:
> What about just showing the two dashes in the example like you have but not adding the text talking about them?

Yes. This can give users a hint on how to use those options properly. I updated the files.

DavidAnson:
Thanks!

PR #415, opened 2023-07-31
add `--verbose` flag

MFTabriz:
`--verbose` flag can be used to ensure linter is actually parsing the files when using globs or for general debugging.

PS: I couldn’t think of a simple solution to logs the relevant messages after each file’s name.


MFTabriz:
The failing test (failing to parse error message on windows) seems to be unrelated to this PR.

MFTabriz:
I think I fixed the path issue on windows!

DavidAnson:
Do you mind applying that test path fix to the pending Dependabot PR? It should pass after that and I'll accept the PR.

markdownlint.js
```diff
@@ -192,6 +192,7 @@ program
   .option('-q, --quiet', 'do not write issues to STDOUT')
   .option('-r, --rules  [file|directory|glob|package]', 'include custom rule files', concatArray, [])
   .option('-s, --stdin', 'read from STDIN (does not work with files)')
+  .option('-v, --verbose', 'write filenames to STDOUT')
```
DavidAnson, line 195: "file names"

DavidAnson (changes requested):
Just a tiny change request and also one thing to think about: Is "verbose" the right name for this flag? It's not wrong but it's also pretty specific about reporting "progress" - so is something like that maybe better?

DavidAnson:
Appreciate the tests, BTW!

markdownlint.js
MFTabriz: done in 414f8acfa2bec0318816bf8c4a37ff58e1553e8e!

DavidAnson (approved):

DavidAnson:
Looks good - tho I can't merge it because it's a Draft PR?

MFTabriz:
> Looks good - tho I can't merge it because it's a Draft PR?

I'm trying an idea (and simultaneously learning a bit of JS). I'll finalize it in a couple of days.

DavidAnson:
I'm not sure there's a lot of room to improve upon what you have here, but I will keep an open mind until I see what you're up to. :)

MFTabriz:
@DavidAnson Thanks for keeping an open mind. 😄I hope you are not allergic to bad JS codes!

Now the `--verbose` flag:

- logs version
- logs the list of the files given to the linter and the list of files reported back separately
- respects --quiet

As you already know, I’m not familiar with idiomatic JS. I’ll appreciate if you can guide me on improving this PR.

markdownlint.js
```diff
@@ -192,11 +218,20 @@ program
...
   .option('-r, --rules  [file|directory|glob|package]', 'include custom rule files', concatArray, [])
   .option('-s, --stdin', 'read from STDIN (does not work with files)')
+  .option('-v, --verbose', 'verbose mode')
   .option('--enable [rules...]', 'Enable certain rules, e.g. --enable MD013 MD041 --')
   .option('--disable [rules...]', 'Disable certain rules, e.g. --disable MD013 MD041 --');
 
 program.parse(process.argv);
 
+if (options.quiet && options.verbose) {
+  options.verbose = false;
```
DavidAnson, line 228: This makes me feel like it's one setting with three values?

markdownlint.js
```diff
@@ -256,6 +291,9 @@ const files = prepareFileList(program.args, ['md', 'markdown']).filter(value =>
 const ignores = prepareFileList(options.ignore, ['md', 'markdown'], files);
 const customRules = loadCustomRules(options.rules);
 const diff = files.filter(file => !ignores.some(ignore => ignore.absolute === file.absolute)).map(paths => paths.original);
+if (options.verbose && !options.stdin) {
+  console.log('files to check:', diff.join(' '));
```
DavidAnson, line 295: This will be all on the same line? That is harder to scan visually.

markdownlint.js
```diff
@@ -299,6 +337,10 @@ function lintAndPrint(stdin, files) {
       const fixResult = markdownlint.sync(fixOptions);
       const fixes = fixResult[file].filter(error => error.fixInfo);
       if (fixes.length > 0) {
+        if (options.verbose && !options.output) {
+          console.log('fixing', file);
```
DavidAnson, line 341: If there is a colon above, let's use one here also.

markdownlint.js
```diff
@@ -116,48 +116,64 @@ function prepareFileList(files, fileExtensions, previousResults) {
...
-        column: (result.errorRange && result.errorRange[0]) || 0,
-        names: result.ruleNames.join('/'),
-        description: result.ruleDescription + (result.errorDetail ? ' [' + result.errorDetail + ']' : '') + (result.errorContext ? ' [Context: "' + result.errorContext + '"]' : '')
-      };
+      return null;
     })
-  );
+    .filter(Boolean);
 
   let lintResultString = '';
```
DavidAnson, line 150: Instead of keeping this, I think it can be removed and recreated if it is needed. It looks like it is empty exactly when the strings array has no elements?

markdownlint.js
```diff
@@ -116,48 +116,64 @@ function prepareFileList(files, fileExtensions, previousResults) {
...
-    }
+        }
 
-    // Note: process.exit(1) will end abruptly, interrupting asynchronous IO
-    // streams (e.g., when the output is being piped). Just set the exit code
-    // and let the program terminate normally.
-    // @see {@link https://nodejs.org/dist/latest-v8.x/docs/api/process.html#process_process_exit_code}
-    // @see {@link https://github.com/igorshubovych/markdownlint-cli/pull/29#issuecomment-343535291}
-    process.exitCode = exitCodes.lintFindings;
+        return `${result.file}: ✔`;
```
DavidAnson, line 173: I would expect failing files to end with a little 'x'?

test/test.js
```diff
@@ -824,3 +824,79 @@ test('configuration can be .cjs in the ESM (module) workspace', async t => {
...
+test('--verbose flag with --fix for correct file', async t => {
+  const result = await execa('../markdownlint.js', ['--fix', '--verbose', 'correct.md'], {stripFinalNewline: false});
+  t.true(result.stdout.includes('files to check: correct.md'));
+  t.true(result.stdout.includes('correct.md: ✔'));
+  t.true(!result.stdout.includes('fixing correct.md'));
+  t.is(result.stderr, '');
+  t.is(result.exitCode, 0);
+});
+
+test('--verbose flag with --fix for incorrect file', async t => {
```
DavidAnson, line 845: Missing test for verbose with broken file (no fix).

test/test.js
```diff
@@ -824,3 +824,79 @@ test('configuration can be .cjs in the ESM (module) workspace', async t => {
...
+test('--stdin and --verbose with valid input logs stdin as source', async t => {
+  const input = ['# Heading', '', 'Text', ''].join('\n');
+  const result = await execa('../markdownlint.js', ['--verbose', '--stdin'], {input, stripFinalNewline: false});
+  t.true(result.stdout.includes('stdin: ✔'));
+  t.true(!result.stdout.includes('files to check:'));
+  t.is(result.stderr, '');
+  t.is(result.exitCode, 0);
+});
+
+test('--output and --verbose with valid input logs nothing to console', async t => {
```
DavidAnson, line 880: I could argue that verbose should always mean "verbose to console". If so, these two test cases would not be different. Just like I don't like how `quiet` and `verbose` overlap, disabling `verbose` for `output` seems not obvious?

DavidAnson (changes requested):
OK, I see what you're going for and it makes sense in the context of what I've seen from other tools! However, the new behavior raises some more questions and I've left new comments. I will need to look at this again on a bigger screen than my phone, but I think this captures all the significant feedback. Thank you!

Issue #113, opened 2020-10-16
Feature: Support reading configuration from pyproject.toml


lyz-code:
Hi. It seems that Python is slowly migrating to store all project configuration into the `pyproject.toml` file ([PEP 518](https://www.python.org/dev/peps/pep-0518/)). 

Can markdownlint-cli support this configuration method? Specifically, it would be nice if markdownlint-cli could understand `[tool.markdownlint]` section in the `pyproject.toml` file.

Not being Python based I can understand that you don't want to invest the effort.

adam-grant-hendry:
+1 Please support!

odimko:
Hi @DavidAnson ! Are there any updates on this? To be able to store `markdownlint-cli`'s configuration in `pyproject.toml` file, like other tools `ruff`, `pytest`, etc do?

DavidAnson:
There are two parts to achieving this. The first is done by the issue linked above. The second is described by me here: https://github.com/igorshubovych/markdownlint-cli/pull/457#issuecomment-1945283551

ModeSevenIndustrialSolutions:
+1 Adding my voice to those requesting support for configuration from pyproject.toml

PR #457, opened 2024-02-13
add possibility to parse toml files like pyproject.toml

BenniWi:
add functionality to read in *.toml files like the pyproject.toml as requested in #113. 

further: add small fix to js and json config file


DavidAnson:
JSON and YAML configuration files store everything in the root-most object. This PR proposes that TOML files would be required to store everything in an object twice nested (toml.markdownlint). I understand this is being done for the sake of an external tool, but it is inconsistent for users of this CLI. Note that CLI2 supports configuration embedded in Node's package.json, but does so as an additional format and not as a requirement for all JSON files. I suspect that won't be enough for you and the folks who opened the corresponding issue, so perhaps it would make sense to add two new entries to the parsers array: one that behaves like you've done here and one that looks for the object at the root? I don't love my proposal, and am open to other ideas.

BenniWi:
I see your point but agree, that checking for both might not be the best solution.
An alternative approach would be to have an additional command line argument to specify in which level to search for the configuration data !?
What do you think about that?

DavidAnson:
I wasn't sure at first, but this idea grew on me when I realized that what you propose would work for any of the configuration file formats! Specifically, your proposal could also be used to enable storing settings inside Node's package.json (similar to what CLI2 allows) - with no additional work.

So my suggestion is that configuration is always assumed to live at the root of all configuration objects (as is already the case today) and you add the new parameter you propose which optionally provides a path into that object for scenarios like yours.

I propose using the JSON path syntax described here: [https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-01.html](https://www.ietf.org/archive/id/draft-ietf-jsonpath-base-01.html). However, I do not think we need to support the full syntax at this point. Instead, I propose the only syntax that should be allowed is "$.first.second" for an arbitrary nesting. "$" alone should be the default and maps to today's behavior. A simple string split on "." should be all the implementation that's needed for now which keeps this simple but leaves the door open for more sophistication in the future.

This makes the proposed change much more work – especially around new test cases. I understand if you're not ready to get into that now, but favor this as the path forward barring other/better suggestions.

DavidAnson:
Cool, good luck! I tend to be very picky, so let me warn you and apologize in advance. :)

If you'd like to post updates as you make progress, I'd be happy to have a look at those and make sure we stay on the same page. It's a lot easier than waiting till the very end and finding out we had completely different ideas.

Thinking briefly about the work here, I don't think there is a lot of new code. You can use the existing command-line parser to add a new single-value parameter. You already found how to add the new TOML parser. The big thing will be test cases that cover the new behavior for the various formats.

One approach you might take - and I might even suggest - is to add the TOML parser in a single self-contained commit. That's basically the work you have already done except that you'd nest everything at the root and remove the nesting. Otherwise, I think what you already have here is close to ready and we could complete this PR for the feature "add TOML parser". Then you can do the work to add the new command-line parameter in a second PR.

It can help to split things so there are fewer moving pieces. Just a thought!

BenniWi:
I switched everything to root now and raised the new issue #458 to add the additional command line argument.

Unfortunately I struggle a bit with the toml parser in general. 
The configuration is correctly parsed, but in the further process, the linting is not done at all. I couldn't figure out the problem so far. 
I observed, that for json/yaml parsers, the config variable contains a prototype object, not only the configuration itself. For toml this is not the case. Don't know if this is somehow related.

README.md
```diff
@@ -102,6 +102,14 @@ A sample configuration file:
 }
 ```
 
+```toml
```
DavidAnson, line 105: I don't think TOML is common enough to get a dedicated section - especially when YAML doesn't. What about removing this and adding a link to one of the TOML test files on the following line that links to examples?

markdownlint.js
```diff
@@ -36,15 +36,19 @@ function yamlParse(text) {
...
 const exitCodes = {
   lintFindings: 1,
   failedToWriteOutputFile: 2,
   failedToLoadCustomRules: 3,
   unexpectedError: 4
 };
 
-const projectConfigFiles = ['.markdownlint.jsonc', '.markdownlint.json', '.markdownlint.yaml', '.markdownlint.yml'];
-const configParsers = [jsoncParse, yamlParse];
+const projectConfigFiles = ['.markdownlint.jsonc', '.markdownlint.json', '.markdownlint.yaml', '.markdownlint.yml', '.markdownlint.toml'];
```
DavidAnson, line 50: I'd prefer not to introduce TOML as an officially supported/searched-for configuration file name because it is NOT supported by CLI2 or the VS Code extension.

package.json
```diff
@@ -45,7 +45,8 @@
     "jsonc-parser": "~3.2.1",
     "markdownlint": "~0.33.0",
     "minimatch": "~9.0.3",
-    "run-con": "~1.3.2"
+    "run-con": "~1.3.2",
+    "toml": "^3.0.0"
```
DavidAnson, line 49: Please match the `~` prefix used by this project for version references.

test/config-files/toml/heading-dollar.md
```diff
@@ -0,0 +1,3 @@
+# Heading$
```
DavidAnson, line 1: Nice, thanks for matching the pattern here!

test/md043-config.js
```diff
@@ -6,6 +6,6 @@
 module.exports = {
   MD012: false,
   MD043: {
-    headers: ['# First', '## Second', '### Third']
+    headings: ['# First', '## Second', '### Third']
```
DavidAnson, line 9: Unrelated, but thanks for the update!

test/config-files/toml/.markdownlint.toml
```diff
@@ -0,0 +1,2 @@
+[no-trailing-punctuation]
```
DavidAnson, line 1: See my comment elsewhere about not introducing `.markdownlint.toml`, please.

test/test.js
```diff
@@ -441,6 +448,8 @@ test('.markdownlint.jsonc in cwd is used automatically', getCwdConfigFileTest('j
 
 test('.markdownlint.json in cwd is used automatically', getCwdConfigFileTest('json'));
 
+test('.markdownlint.toml in cwd is used automatically', getCwdConfigFileTest('toml'));
```
DavidAnson, line 451: This test should be removed if we don't introduce `.markdownlint.toml`.

DavidAnson (changes requested):
Everything here looks like I expected. I've left a few comments - only one real point of discussion.

That said, can you please explain your comment, "Unfortunately I struggle a bit with the toml parser in general. ..."? I don't understand the problem - as I said, everything here looks like I expected. What isn't working for TOML that works for JSON/YAML?

test/test.js
DavidAnson: Oh, this test fails. Is that the problem you mentioned? Moot if we decide to remove it, though it'd be good to understand why.

DavidAnson: This is why it fails: https://github.com/BinaryMuse/toml-node/issues/55

Here's a commit to fix it that Codespaces won't let me push (you can apply it directly):

```diff
commit 0a1e1bdd6a58526db1bca0c4f02d54973b425d1d (HEAD -> add_toml_parser)
Author: David Anson <david@dlaa.me>
Date:   Mon Feb 19 23:06:57 2024 +0000

    Fix toml.parse issue of returning null-prototype objects.

diff --git a/markdownlint.js b/markdownlint.js
index 108add7..78be74c 100755
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -37,7 +37,7 @@ function yamlParse(text) {
 }
 
 function tomlParse(text) {
-  return require('toml').parse(text);
+  return require('deep-extend')({}, require('toml').parse(text));
 }
 
 const exitCodes = {
```

DavidAnson: Please make the change above regardless. :)

README.md
BenniWi: I see your point and removed the section.
Linking to the test example is not the best solution since the other examples given are way more complete. I converted the "base-config.json" file to toml and added it to the test folder to have an example to link.

markdownlint.js
BenniWi: Fair point. I removed it.

package.json
BenniWi: Changed it. 
Nevertheless it seems to be inconsistent since the devDependencies are using ^

test/config-files/toml/.markdownlint.toml
BenniWi: Removed the file

test/test.js
BenniWi: Thanks for the fix, would have taken me several days to figure that out :bow: .

Test is removed

package.json
DavidAnson: I reference all dependencies exactly in my projects; this one is more strict about dependencies than devDependencies.

DavidAnson (approved):
Thanks for the updates!

package.json
BenniWi: Ok, makes sense. Thanks for the clarification

BenniWi:
Hi @DavidAnson, thanks a lot for your time and for guiding me through this. 
It is my first time working with JS and I really appreciate your help.

Issue #508, opened 2024-12-05
Incompatible dependency: glob requires Node 20+


bmish:
markdownlint-cli depends on glob.

glob v11 requires Node 20+: https://github.com/isaacs/node-glob/blob/main/changelog.md#110

But markdownlint-cli currently specifies that it supports Node >= 18:

https://github.com/igorshubovych/markdownlint-cli/blob/0d9fcb51a54f3b750b911c054b4bd1a590f1b592/package.json#L10C16-L10C18

So the solution would be to downgrade glob until requiring Node 20+.

DavidAnson:
True, but that was 6 months ago and this doesn't seem to have been a problem for CI or users. (I wish Dependabot wouldn't offer updates like this that are semi-incompatible.)

DavidAnson:
@bmish, do you know of a tool that validates this (and could run in CI) or did you just happen to spot this manually?

bmish:
Here's a fix that would catch it:

* https://github.com/igorshubovych/markdownlint-cli/pull/509 

I will note that this isn't really an urgent problem, so I don't mind if you want to:
* Downgrade glob
* Or just wait until [2025-04-30](https://github.com/nodejs/Release#release-schedule) to drop support for Node 18 and merge the fix so it doesn't happen again

DavidAnson:
My current thinking is to wait, but leave the issue open for visibility. :)

DavidAnson:
@bmish, FYI that markdownlint library and CLI2 now both have engine-strict enabled (without issue on Node 18) in their "next" branches.

lsh-0:
> My current thinking is to wait, but leave the issue open for visibility. :)

If we could get a 0.44.0 with a downgraded `glob` that would be helpful, thank you

DavidAnson:
@lsh-0, versions `0.41.0` and earlier of `markdownlint-cli` use the older version of `glob` that declares support for Node `18`. Node `20` and later are supported by `glob` with the latest version of `markdownlint-cli`. (And there have been no reported issues due to this in 6 months.) Is there a compelling case to downgrade?

lsh-0:
> Is there a compelling case to downgrade?

The set of dependencies described by `markdownlint-lint` are incompatible. It doesn't support node18 any more because of `glob@11.x`.

Downgrading `glob` seems the most natural fix but I guess you could also change this line and remove support for node18:

[...]

PR #526, opened 2025-02-25
chore: switch to tinyglobby

benmccann:
https://npmgraph.js.org/?q=glob@10.4.5 - 26 dependencies
https://npmgraph.js.org/?q=tinyglobby - 2 dependencies

It also supports much older versions of Node avoiding the issues in https://github.com/igorshubovych/markdownlint-cli/issues/508. Because we can upgrade to the latest version it will also dedupe better whereas users are split across multiple majors for `glob`

### DIFF
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -6,7 +6,7 @@ import Module from 'node:module';
 import os from 'node:os';
 import process from 'node:process';
 import {program} from 'commander';
-import {glob} from 'glob';
+import {globSync} from 'tinyglobby';
 import {applyFixes} from 'markdownlint';
 import {lint, readConfig} from 'markdownlint/sync';
 import rc from 'run-con';
@@ -93,7 +93,8 @@ function readConfiguration(userConfigFile) {
 function prepareFileList(files, fileExtensions, previousResults) {
   const globOptions = {
     dot: Boolean(options.dot),
-    nodir: true
+    onlyFiles: true,
+    expandDirectories: false
   };
   let extensionGlobPart = '*.';
   if (!fileExtensions) {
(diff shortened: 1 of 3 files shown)


benmccann:
While true that glob is more popular, tinyglobby is popular enough that any bugs would quickly be discovered. If you compare tinyglobby to any other library, it's still popular on its own. tinyglobby's usage increased 50% in the past month, so it's being adopted pretty quickly.

DavidAnson:
There are a couple of open issues that seem alarming. There is no rush here, so I'm inclined to give this some more time to bake.

DavidAnson:
Could you please update the PR to address the merge conflicts and then I will approve CI to run?

benmccann:
Done. Thanks!

DavidAnson:
Okay, thanks, I'm good to give this a try. Can you please convince me the changes you made to `globOptions` are the most consistent possible? Did you review all the previous defaults and new settings? How do we make sure we're not accidentally enabling some troublesome new behavior like enumerating dotfiles (or whatever)?

benmccann:
Thanks for taking the time to think this change through and point out the chance that we don't get the options mapped one-to-one. I just did another review and pushed one to maintain case sensitivity the same way `glob` does. The rest look the same to me. Ideally the tests can catch regressions for us. I do see the most common cases covered there like globbing for `**/*.md`.

DavidAnson:
GitHub makes this workflow annoying, so I applied your changes directly: https://github.com/igorshubovych/markdownlint-cli/commit/ef3703327d352098f96a38d921fd9d1cf0e39c14

Thanks!

benmccann:
thanks!!

Issue #304, opened 2022-07-16
Unable to pipe output


txtsd:
I'm trying to `grep` for specific rules in the output, but it does not work.

I want to be able to run `markdownlint . | grep MD025` for example.

DavidAnson:
Errors are written to `stderr`, not `stdout`. Your shell probably has a way to change what is piped.

txtsd:
Thanks! I found it.

POSIX compliant:

```bash
markdownlint . 2>&1 | grep MD025
```

ZSH specific:

```zsh
markdownlint . |& grep MD025
```

PR #649, opened 2026-08-07
Say what happened when the given files match nothing

### LINKED ISSUE #104: Misleading error message when no input files are detected
Continued from [markdownlint#304](https://github.com/DavidAnson/markdownlint/issues/304)

```sh
markdownlint --fix '**/*.md'
```

When running this command in a project without any md files, it gives me confusing error:

```sh
  Usage: markdownlint [options] <files|directories|globs>

  MarkdownLint Command Line Interface

  Options:

    -h, --help                                  output usage information
    -V, --version                               output the version number
    -f, --fix                                   fix basic errors (does not work with STDIN)
    -s, --stdin                                 read from STDIN (does not work with files)
    -o, --output [outputFile]                   write issues to file (no console)
    -c, --config [configFile]                   configuration file (JSON, JSONC, JS, or YAML)
    -i, --ignore [file|directory|glob]          file(s) to ignore/exclude
    -p, --ignore-path [file]                    path to file with ignore pattern(s)
    -r, --rules  [file|directory|glob|package]  custom rule files
```

It was my mistake to forget to add a readme file, but I think in this case the output should be something more specific, like `No valid files found`. Now it suggests I entered invalid command.

Eljees:
Fixes #104.

A glob or path that resolves to no Markdown file leaves `inputs` empty — the same state as being given no arguments at all — so the dispatch at the bottom of `markdownlint.js` falls through to `program.help()`:

```console
$ cd a-project-with-no-markdown
$ markdownlint --fix '**/*.md'
Usage: markdownlint [options] [files|directories|globs...]
...
```

That output is indistinguishable from running `markdownlint` with no arguments, so it reads as "you typed the command wrong" when the command was fine and the directory simply held no Markdown. That's what @paulmelnikow ran into, and #304 upstream.

The new branch is taken only when operands were actually given and `--stdin` wasn't, so the three cases that legitimately show usage — no arguments, files with `--stdin`, `--fix` with `--stdin` — keep doing so, and their tests still pass unchanged.

The message goes to stderr, which leaves `--output`, `--json` and `--quiet` consumers of stdout untouched. **The exit status is unchanged (0)**: a build that lints a tree which happens to contain no Markdown passes today and still passes, so this isn't a breaking change for anyone's CI.

Two tests cover it — a glob that matches nothing and a path to a file that doesn't exist. Both fail on current `master` (they get the help text) and pass with the change. The full suite is green at 91 tests, and `xo` is clean.

AI-assisted (LLM used for drafting); the runs above are mine.

### DIFF
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -339,6 +339,11 @@ try {
     lintAndPrint(null, diff);
   } else if (inputs.length === 0 && options.stdin && !options.fix) {
     import('node:stream/consumers').then(module => module.text(process.stdin)).then(lintAndPrint);
+  } else if (program.args.length > 0 && !options.stdin) {
+    // Files, directories, or globs were given, but none of them resolved to a
+    // Markdown file. That is not a usage error, so say what happened instead
+    // of printing the help text and leaving the user to guess.
+    console.error('No matching files found.');
   } else {
     program.help();
   }
--- a/markdownlint.js
+++ b/markdownlint.js
@@ -340,10 +340,8 @@ try {
   } else if (inputs.length === 0 && options.stdin && !options.fix) {
     import('node:stream/consumers').then(module => module.text(process.stdin)).then(lintAndPrint);
   } else if (program.args.length > 0 && !options.stdin) {
-    // Files, directories, or globs were given, but none of them resolved to a
-    // Markdown file. That is not a usage error, so say what happened instead
-    // of printing the help text and leaving the user to guess.
-    console.error('No matching files found.');
+    // Files, directories, or globs were given, but none of them resolved to a Markdown file.
+    console.warn('No matching files found');
   } else {
     program.help();
   }
--- a/test/test.js
+++ b/test/test.js
@@ -44,17 +44,24 @@ test('no files shows help', async t => {
   t.is(result.exitCode, 0);
 });
 
-test('globs that match nothing report that, not usage', async t => {
-  const result = await spawn('../markdownlint.js', ['no-such-directory/**/*.md']);
+test('a missing file reports that', async t => {
+  const result = await spawn('../markdownlint.js', ['no-such-file.md']);
   t.is(result.stdout, '');
-  t.is(result.stderr, 'No matching files found.');
+  t.is(result.stderr, 'No matching files found');
   t.is(result.exitCode, 0);
 });
 
-test('a missing file reports that, not usage', async t => {
-  const result = await spawn('../markdownlint.js', ['no-such-file.md']);
+test('a missing directory reports that', async t => {
+  const result = await spawn('../markdownlint.js', ['no-such-directory']);
   t.is(result.stdout, '');
-  t.is(result.stderr, 'No matching files found.');
+  t.is(result.stderr, 'No matching files found');
+  t.is(result.exitCode, 0);
+});
+
+test('globs that match nothing report that', async t => {
+  const result = await spawn('../markdownlint.js', ['no-such-directory/**/*.md']);
+  t.is(result.stdout, '');
+  t.is(result.stderr, 'No matching files found');
   t.is(result.exitCode, 0);
 });
 
--- a/test/test.js
+++ b/test/test.js
@@ -44,6 +44,20 @@ test('no files shows help', async t => {
   t.is(result.exitCode, 0);
 });
 
+test('globs that match nothing report that, not usage', async t => {
+  const result = await spawn('../markdownlint.js', ['no-such-directory/**/*.md']);
+  t.is(result.stdout, '');
+  t.is(result.stderr, 'No matching files found.');
+  t.is(result.exitCode, 0);
+});
+
+test('a missing file reports that, not usage', async t => {
+  const result = await spawn('../markdownlint.js', ['no-such-file.md']);
+  t.is(result.stdout, '');
+  t.is(result.stderr, 'No matching files found.');
+  t.is(result.exitCode, 0);
+});
+
 test('files and --stdin shows help', async t => {
   const result = await spawn('../markdownlint.js', ['--stdin', 'correct.md']);
   t.true(result.stdout.includes('--help'));


royaldark:
I was just going down the rabbit hole investigating this exact issue when I saw your MR. Talk about good timing.

Should the behavior differ based on whether the provided inputs are concrete files, versus whole directories?

We do a lot of runs with `markdownlint .` -- and it'd be nice in that case if it just exited 0 silently, even if no files in that directory matched.

If running over specific files, like `markdownlint README.md`, then failing loudly makes good sense.

Eljees:
Good timing indeed. Two things worth separating.

First, a correction on what the PR does: it doesn't touch the exit status. It was 0 in every one of these cases before and it still is. The only change is that `No matching files found.` goes to stderr instead of the usage block. So `markdownlint .` in a tree with no Markdown already exits 0 today, and still will — it just prints one line rather than the help text.

Second, your distinction is implementable. `prepareFileList` already tells the cases apart: it does `fs.lstatSync(file).isDirectory()` to decide whether to expand an operand as a directory. So the message can be suppressed when every operand is an existing directory, and kept when at least one operand is a concrete path or a glob that resolved to nothing.

Two things I'd want your read on before I change it:

- A glob that matches nothing (`markdownlint '**/*.md'`) is the case in #104. I'd keep the message there, since the glob is explicit intent rather than "lint whatever is in this tree". Does that match what you want?
- By "failing loudly" for `markdownlint README.md`, do you mean a non-zero exit when the file doesn't exist? That is a real behaviour change — it passes today — so I'd rather keep it out of this PR and raise it separately than fold it in here.

Eljees:
@royaldark gentle ping — the two questions at the end of my last comment are what this is waiting on. (1) Should a glob that matches nothing keep printing the message even if a bare directory operand suppresses it? That glob case is what #104 is about, so I'd keep the message there. (2) By "failing loudly" for `markdownlint README.md`, did you mean a non-zero exit for a concrete file that doesn't exist? That one is a real behaviour change and I'd rather raise it separately than fold it in here. Confirm (1) and I'll push the directory-suppression variant.

Eljees:
@royaldark — following up on the two questions from last week, and reframing so the PR isn't blocked on them.

Worth separating what's here from what you asked about. This PR doesn't change exit status: `markdownlint .` on a tree with no Markdown exits 0 today and still does afterwards. The only change is that one line goes to stderr instead of the usage block. So the directory-vs-file distinction you want is a behaviour change on top of this, not a correction to it.

That makes the current diff self-contained. Suppressing the message when every operand is an existing directory is a small follow-up — `prepareFileList` already tells the cases apart via `lstatSync().isDirectory()` — and I'll open it separately if you still want it. My reading is that a glob matching nothing should keep the message either way, since a glob is explicit intent rather than "lint whatever is in this tree", but that's the part I'd want your read on.

@igorshubovych green on all 9 checks, mergeable, no conflicts. Is there anything you'd want changed before this can go in?

DavidAnson:
@Eljees, I was waiting to see how the discussion above was resolved. I agree the new behavior seems clearer and that putting in a special case for directories seems unnecessary. I'll review the changes soon.

test/test.js
```diff
@@ -44,6 +44,20 @@ test('no files shows help', async t => {
   t.is(result.exitCode, 0);
 });
 
+test('globs that match nothing report that, not usage', async t => {
```
DavidAnson: Please remove ", not usage" - there's no need to document behavior that's not present.

DavidAnson: (Here and below)

test/test.js
```diff
@@ -44,6 +44,20 @@ test('no files shows help', async t => {
...
+  const result = await spawn('../markdownlint.js', ['no-such-directory/**/*.md']);
+  t.is(result.stdout, '');
+  t.is(result.stderr, 'No matching files found.');
+  t.is(result.exitCode, 0);
+});
+
```
DavidAnson, line 53: Please add a third test here for the 'no-such-directory' scenario explicitly.

markdownlint.js
```diff
@@ -339,6 +339,11 @@ try {
     lintAndPrint(null, diff);
   } else if (inputs.length === 0 && options.stdin && !options.fix) {
     import('node:stream/consumers').then(module => module.text(process.stdin)).then(lintAndPrint);
+  } else if (program.args.length > 0 && !options.stdin) {
+    // Files, directories, or globs were given, but none of them resolved to a
+    // Markdown file. That is not a usage error, so say what happened instead
+    // of printing the help text and leaving the user to guess.
+    console.error('No matching files found.');
```
DavidAnson: Remove ".", please - other messages do not end with punctuation.

test/test.js
DavidAnson: As a nitpick, I'd order tests in the order: file, directory, glob.

markdownlint.js
```diff
@@ -339,6 +339,11 @@ try {
     lintAndPrint(null, diff);
   } else if (inputs.length === 0 && options.stdin && !options.fix) {
     import('node:stream/consumers').then(module => module.text(process.stdin)).then(lintAndPrint);
+  } else if (program.args.length > 0 && !options.stdin) {
+    // Files, directories, or globs were given, but none of them resolved to a
```
DavidAnson: First sentence is good, please collapse to a single line - second sentence is unnecessary, please remove.

markdownlint.js
DavidAnson: Because this is considered a "success" case, I feel `console.warn` is more appropriate - but maybe there's a reason to use `error` instead?

DavidAnson (changes requested):
Thank you! I left a few comments/questions.

test/test.js
Eljees: Removed from both test names in 7a68e7c.

test/test.js
Eljees: Added `a missing directory reports that` in 7a68e7c, and reordered the three as file, directory, glob.

I checked the bare-directory case actually reaches this branch before writing the test for it: `node ../markdownlint.js no-such-directory` from `test/` prints the message on stderr and exits 0, same as the missing file and the glob.

markdownlint.js
Eljees: Dropped in 7a68e7c - the message is now `No matching files found`.

markdownlint.js
Eljees: Collapsed to one line and removed the second sentence in 7a68e7c.

markdownlint.js
Eljees: No reason to prefer `error` - switched to `console.warn` in 7a68e7c. `warn` is already used at markdownlint.js:196, so it is not a new idiom here, and both write to stderr in Node, so the exit code and the `result.stderr` assertions in the tests are unchanged.

DavidAnson (approved):
Thank you for the quick changes!

